PR #6, opened 2018-04-13
Fix go test error and optimize the verify block method: reduce the lock count of the transaction pool

lightshine001:

### DIFF
--- a/txnpool/common/transaction_pool_test.go
+++ b/txnpool/common/transaction_pool_test.go
@@ -21,7 +21,6 @@ package common
 import (
 	"bytes"
 	"encoding/hex"
-	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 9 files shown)


txnpool/common/transaction_pool_test.go
```diff
@@ -78,11 +78,15 @@ func TestTxPool(t *testing.T) {
...
 
-	txList := txPool.GetTxPool(true)
+	txList, oldTxList := txPool.GetTxPool(true, 0)
 	for _, v := range txList {
 		fmt.Println(v)
 	}
 
+	for _, v := range oldTxList {
+		fmt.Println(v)
+	}
```
laizy: use `assert` to replace all those `println` stuff

laizy (changes requested):

txnpool/common/transaction_pool_test.go
lightshine001: Udpated.

laizy (approved):

PR #21, opened 2018-04-16
Net2: new network interface and process

BooniesFX:
1.  new p2p interface
2.  message router register  
3.  logic layer more clear


p2pserver/peer/config.json
```diff
@@ -0,0 +1,30 @@
+﻿
```
lightshine001: Duplicated. Please remove it.

BooniesFX: After default config fix, the file is useless . I will remove it

p2pserver/message/msg_pack/msg_pack.go
```diff
@@ -0,0 +1,491 @@
...
+func NewBlock(bk *ct.Block) ([]byte, error) {
+	log.Debug()
+	var blk mt.Block
+	blk.Blk = *bk
+
+	tmpBuffer := bytes.NewBuffer([]byte{})
+	bk.Serialize(tmpBuffer)
+
+	p := new(bytes.Buffer)
+	err := binary.Write(p, binary.LittleEndian, tmpBuffer.Bytes())
```
laizy: serialize a []byte is meaningless. use `tmpBuffer` directly

p2pserver/message/msg_pack/msg_pack.go
```diff
@@ -0,0 +1,491 @@
...
+	if err != nil {
+		log.Error("Binary Write failed at new Msg")
+		return nil, err
+	}
+
+	checkSumBuf := mt.CheckSum(p.Bytes())
+	blk.Init("block", checkSumBuf, uint32(len(p.Bytes())))
+	log.Debug("The message payload length is ", blk.Length)
+
+	m, err := blk.Serialization()
```
laizy: double serialization is wasteful

laizy (changes requested):

p2pserver/message/msg_pack/msg_pack.go
BooniesFX: ok, i will delete it

p2pserver/message/msg_pack/msg_pack.go
BooniesFX: refactoring msg logic to reduce useless serialization

PR #23, opened 2018-04-17
fix return nil

xizho10:


Arbio5zt (approved):
approve
1.fixed unknow block query bug.
2.removed unnecessary output log.

PR #24, opened 2018-04-17
ontology docker build support, based on ubuntu-core image

Honglei-Cong:
'make docker' will make docker image for ontology.
After docker image built, "docker run -d <image-id>" will start ontology docker with solo mode.


Honglei-Cong:
to keep platform independent, ontology should be built from docker, rather than copy native ontology bin into docker image.

PR #76, opened 2018-04-26
refactor: make sure that interface has been implemented completely by entity

Yihen-Liu:
Static (compile time) check that each entity satisfies the interface.


laizy:
It seems meaningless to me.

PR #82, opened 2018-04-26
optimize block sync and store

gasby88:

### DIFF
--- a/net/message/blockHdr.go
+++ b/net/message/blockHdr.go
@@ -283,7 +283,7 @@ func GetHeadersFromHash(startHash common.Uint256, stopHash common.Uint256) ([]ty
 		headers = append(headers, *hd)
 	}
 
-	return headers, count, nil
+	return headers, uint32(len(headers)), nil
 }
 
 func NewHeaders(headers []types.Header, count uint32) ([]byte, error) {
--- a/net/message/blockHdr.go
+++ b/net/message/blockHdr.go
@@ -24,13 +24,13 @@ import (
 	"encoding/binary"
 	"errors"
 
+	"fmt"
 	"github.com/ontio/ontology/common"
 	"github.com/ontio/ontology/common/log"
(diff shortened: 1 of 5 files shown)


net/message/blockHdr.go
```diff
@@ -283,7 +283,7 @@ func GetHeadersFromHash(startHash common.Uint256, stopHash common.Uint256) ([]ty
 		headers = append(headers, *hd)
 	}
 
-	return headers, count, nil
+	return headers, uint32(len(headers)), nil
```
laizy: return the length of headers is redundant , need adjust the function signature

net/node/block_sync.go
```diff
@@ -355,11 +361,16 @@ func (this *BlockSyncMgr) OnBlockReceive(block *types.Block) {
...
+		log.Errorf("BlockSyncMgr syncBlock GetCurrentBlockHeight error:%s", err)
+		return
+	}
+	if height <= curBlockHeight {
+		return
+	}
+	this.addBlockCache(block)
+	go this.saveBlock()
 	go func() {
 		time.After(time.Millisecond * 10)
```
laizy: since background Ticker will fire per second , this seems redundant.

net/node/block_sync.go
```diff
@@ -392,104 +403,125 @@ func (this *BlockSyncMgr) OnDelNode(nodeId uint64) {
 	log.Infof("BlockSyncMgr OnDelNode:%d", nodeId)
 }
 
-func (this *BlockSyncMgr) isSyncingBlock() bool {
+func (this *BlockSyncMgr) syncingHeader() bool {
```
laizy: `tryGetSyncHeaderLock` and `releaseSyncHeaderLock`  is a better name?

laizy (changes requested):

laizy (approved):

PR #84, opened 2018-04-27
ontology global param config native smart contract

qiluge:


luodanwg:
need a version control of global param change

Arbio5zt (dismissed):
need a version control of global param change

PR #93, opened 2018-05-02
Bypass vbft msg and  Remove peer which have error occur in connecting state.

BooniesFX:
Bypass vbft msg actor under other consensus algo
Remove peer which have error occur in connecting state.

Signed-off-by: Xiang Fu <fuxiang@onchain.com>

### DIFF
--- a/p2pserver/message/utils/msg_handler.go
+++ b/p2pserver/message/utils/msg_handler.go
@@ -244,12 +244,14 @@ func VersionHandle(data *msgCommon.MsgPayload, p2p p2p.P2P, pid *evtActor.PID, a
 	if version.P.IsConsensus == true {
 		if config.DefConfig.P2PNode.DualPortSupport == false {
 			log.Warn("consensus port not surpport")
+			p2p.RemoveFromConnectingList(data.Addr)
 			return errors.New("consensus port not surpport")
 		}
 		remotePeer := p2p.GetPeerFromAddr(data.Addr)
 
 		if remotePeer == nil {
 			log.Warn(" peer is not exist", data.Addr)
+			p2p.RemoveFromConnectingList(data.Addr)
 			return errors.New("peer is not exist ")
 		}
 		p := p2p.GetPeer(version.P.Nonce)
@@ -257,6 +259,7 @@ func VersionHandle(data *msgCommon.MsgPayload, p2p p2p.P2P, pid *evtActor.PID, a
 		if p == nil {
 			log.Warn("sync link is not exist", version.P.Nonce)
 			p2p.RemovePeerConsAddress(data.Addr)
+			p2p.RemoveFromConnectingList(data.Addr)
 			return errors.New("sync link is not exist")
 		} else {
 			//p synclink must exist,merged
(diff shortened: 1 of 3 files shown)


p2pserver/message/utils/msg_handler.go
```diff
@@ -259,12 +262,14 @@ func VersionHandle(data *msgCommon.MsgPayload, p2p p2p.P2P, pid *evtActor.PID, a
...
 			remotePeer.CloseCons()
 			p2p.RemovePeerConsAddress(data.Addr)
+			p2p.RemoveFromConnectingList(data.Addr)
 			return errors.New("The node handshake with itself ")
 		}
 
 		s := remotePeer.GetConsState()
 		if s != msgCommon.INIT && s != msgCommon.HAND {
 			log.Warn("Unknown status to received version", s)
+			p2p.RemoveFromConnectingList(data.Addr)
```
laizy: you may also need close the net connection .

BooniesFX: clean remove logic, close connection contain same action
if dual port not support, set consensus port to zero which avoid be connected by other peer

p2pserver/message/msg_pack/msg_pack.go
```diff
@@ -208,12 +208,15 @@ func NewVersionPayload(n p2pnet.P2P, isCons bool, height uint32) mt.VersionPaylo
...
 		Services:     n.GetServices(),
 		SyncPort:     n.GetSyncPort(),
-		ConsPort:     n.GetConsPort(),
 		Nonce:        n.GetID(),
 		IsConsensus:  isCons,
 		HttpInfoPort: n.GetHttpInfoPort(),
 	}
-
+	if config.DefConfig.P2PNode.DualPortSupport {
+		vpl.ConsPort = n.GetConsPort()
```
Honglei-Cong: why not set ConsPort right on p2p module initialisation, so n.GetConsPort() returns expected answer?

p2pserver/message/utils/msg_handler.go
```diff
@@ -441,16 +436,17 @@ func VerAckHandle(data *msgCommon.MsgPayload, p2p p2p.P2P, pid *evtActor.PID, ar
 				}
 				nodeConsensusAddr := addr[:i] + ":" +
 					strconv.Itoa(int(remotePeer.GetConsPort()))
-				go p2p.Connect(nodeConsensusAddr, true)
+				if remotePeer.GetConsPort() > 0 {
```
Honglei-Cong: change the upper if as:
       if config.DefConfig.P2PNode.DualPortSupport && remotePeer.GetConsPort() > 0

is 'return nil' ok "if i < 0"?
the following p2p.Send(msgpay.NewAddrReq()) is not required?

p2pserver/message/msg_pack/msg_pack.go
BooniesFX: good idea, logic added in netserver

p2pserver/message/utils/msg_handler.go
BooniesFX: refer to https://github.com/ontio/ontology/pull/178, the handler logic would remove error value.so all return error will delete in that pull.

NewAddrReq is needed for acquiring nbr node`s address.
other changes completed in latest code.

laizy (approved):

PR #94, opened 2018-05-02
add unittest  in http ledger_test

lucas7788:
add unittest  in http ledger_test

### DIFF
--- a/http/base/actor/ledger_test.go
+++ b/http/base/actor/ledger_test.go
@@ -1,0 +1,209 @@
+package actor
+
+import (
+	ldgactor "github.com/ontio/ontology/core/ledger/actor"
+	"testing"
+	"github.com/ontio/ontology-eventbus/actor"
+	"github.com/ontio/ontology/common/log"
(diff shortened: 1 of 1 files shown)


http/base/actor/ledger_test.go
```diff
@@ -0,0 +1,209 @@
...
+	txHash := []common.Uint256{}
+	for _, t := range transactions {
+		txHash = append(txHash, t.Hash())
+	}
+	txRoot, _ := common.ComputeMerkleRoot(txHash)
+
+
+	blockRoot := ledger.DefLedger.GetBlockRootWithNewTxRoot(txRoot)
+
+	prevHash,_ := CurrentBlockHash()
```
qiluge, line 134: the function currentBlockHash has one return value in latest version

http/base/actor/ledger_test.go
```diff
@@ -0,0 +1,209 @@
...
+	height, err := BlockHeight()
```
qiluge, line 135: the BlockHeight() and AddBlock() function don't exist in latest version

PR #97, opened 2018-05-03
cli-cmd/account: add new files and change some related packages.

Doresimon:
# modify some functions to unique package.

1. func: password enter
2. func: add account
3. func: set account
4. func: delete account
5. func: encrypt account file

# refactor types from account.go to account_cmd.go

# add -label flag for <account add> <account set>
* <account list> will show label info at the same time
* default label for an account is ""(blank string)

# add <account import> sub command
* account import -source xxx: to import accounts in xxx to current wallet file.
* use -file <filename> to specify current wallet file. default name is "wallet.dat". 

# change some print info.

Signed-off-by: Halo Guan <guanjixing@onchain.com>


cmd/account_cmd.go
```diff
@@ -321,12 +314,12 @@ func encrypt(ctx *cli.Context) error {
 	prv, _ := keypair.DecryptPrivateKey(wallet.Accounts[index-1].GetKeyPair(), oldPass)
 
 	//let user enter password with double check
-	pass := password.DoubleEnterPassword()
+	pass, _ := password.GetConfirmedPassword()
```
laizy: handle all err in this file.

account/account.go
```diff
@@ -106,3 +109,132 @@ func (ac *Account) PubKey() keypair.PublicKey {
...
+	"4": {"P-521", keypair.P521},
+
+	"P-224": {"P-224", keypair.P224},
+	"P-256": {"P-256", keypair.P256},
+	"P-384": {"P-384", keypair.P384},
+	"P-521": {"P-521", keypair.P521},
+
+	"SM2P256V1": {"SM2P256V1", keypair.SM2P256V1},
+	"ED25519":   {"ED25519", keypair.ED25519},
+}
```
laizy: so why move all those maps here? account module does not care any of those stuff. they are just used to map user's  command line input.  so keep it untouched should be a better choice.

laizy (changes requested):

almoslim4 (approved):

almoslim4 (approved):

almoslim4 (approved):

almoslim4 (approved):

PR #131, opened 2018-05-09
ONT-476 Fix bug: handle the response's height from validator 

lightshine001:

### DIFF
--- a/main.go
+++ b/main.go
@@ -198,7 +198,7 @@ func ontMain(ctx *cli.Context) {
 		os.Exit(1)
 	}
 	p2p.SetPID(p2pPID)
-	err = p2p.Start(true)
+	err = p2p.Start()
 	if err != nil {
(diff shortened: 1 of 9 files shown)


txnpool/proc/txnpool_worker.go
```diff
@@ -41,14 +41,20 @@ type pendingTx struct {
 	flag    uint8           // For different types of verification
 	retries uint8           // For resend to validator when time out before verified
 	ret     []*tc.TXAttr    // verified results
+	height  uint32
+}
+
+type reValidateTx struct {
+	height uint32 // To be re-verified at the height
```
laizy: tracking height in each transaction validation is obvious unsound, since when a new block checking request arrive, the old one will be replaced, and the height will change.

laizy (changes requested):

PR #136, opened 2018-05-10
fix txpool verify block offset 1 bug

laizy:


lightshine001 (approved):
dbft-based unit test verified

lightshine001 (approved):

PR #150, opened 2018-05-14
vbft governance support

Honglei-Cong:
1. governance contract support
2. fee splitting for deposited tokens
3. vbft chain support for governance
4. vbft bug fixings


consensus/vbft/block_pool.go
```diff
@@ -623,7 +616,18 @@ func (pool *BlockPool) setBlockSealed(block *Block, forEmpty bool) error {
...
+		emptyAddr := common.Address{}
+		for _, tx := range block.Block.Transactions {
+			if tx.TxType == types.Invoke && bytes.Compare(tx.Payer[:], emptyAddr[:]) == 0 {
+				systemTxs = append(systemTxs, tx)
+			} else {
+				break
+			}
+		}
+
+		block.Block.Transactions = systemTxs
```
laizy: when dealing with `block, transaction` etc, we implicitly assume they are immutable after construction. here you have modified block's transactions, so you need explicitly recalculate transaction root and set block.Hash to nil.

laizy (changes requested):

consensus/vbft/msg_pool.go
```diff
@@ -49,27 +48,19 @@ func newConsensusRound(num uint64) *ConsensusRound {
...
 }
 
-func (self *ConsensusRound) hasMsg(msg ConsensusMsg) (bool, error) {
-	h, err := HashMsg(msg)
-	if err != nil {
-		return false, fmt.Errorf("failed to hash msg: %s", err)
-	}
-	if _, present := self.msgHashs[h]; present {
+func (self *ConsensusRound) hasMsg(msg ConsensusMsg, msgHash common.Uint256) (bool, error) {
+	if _, present := self.msgHashs[msgHash]; present {
```
laizy, line 63: need adjust the return value to `bool`

Honglei-Cong:
Some important fix needed for the pull request.
close it now.

PR #152, opened 2018-05-15
  refactor: delete cli dir and nodectl(#ONT-536)

Yihen-Liu:
There is unused for cli dir and nodectl file ,which has been replaced by cmd dir.
 Signed-off-by: liuyihen@onchain.com


Arbio5zt:
duplicated with #157

PR #160, opened 2018-05-16
  refactor: ontid smartcontract code is disunity with others(#ONT-545)

Yihen-Liu:
Signed-off-by: liuyihen@onchain.com


laizy:
seems fixed in #163

PR #184, opened 2018-05-18
ONT-551 add export account

AlverLyu:
Export accounts to another wallet file. Support re-encryption for lower performance devices.

### DIFF
--- a/account/client.go
+++ b/account/client.go
@@ -72,6 +72,8 @@ type Client interface {
 	ChangePassword(address string, oldPasswd, newPasswd []byte) error
 	//Change sig scheme to account
 	ChangeSigScheme(address string, sigScheme s.SignatureScheme, passwd []byte) error
+	//Get the underlying wallet data
+	GetWalletData() *WalletData
 }
 
 func Open(path string) (Client, error) {
@@ -313,7 +315,9 @@ func (this *ClientImpl) getAccount(accData *AccountData, passwd []byte) (*Accoun
 	if !accData.VerifyPassword(passwd) {
 		return nil, fmt.Errorf("verify password failed")
 	}
+	keypair.SetScryptParam(this.walletData.Scrypt)
 	privateKey, err := keypair.DecryptPrivateKey(&accData.ProtectedKey, passwd)
+	keypair.SetScryptParam(nil)
 	if err != nil {
 		return nil, fmt.Errorf("decrypt PrivateKey error:%s", err)
 	}
@@ -582,4 +586,8 @@ func (this *ClientImpl) checkSigScheme(keyType, sigScheme string) bool {
 		return false
 	}
 	return true
+}
+
+func (this *ClientImpl) GetWalletData() *WalletData {
+	return this.walletData
 }
--- a/account/file_store.go
+++ b/account/file_store.go
@@ -24,12 +24,10 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-
-	"github.com/ontio/ontology-crypto/keypair"
-	//"github.com/ontio/ontology/core/types"
 	"io/ioutil"
 	"os"
 
+	"github.com/ontio/ontology-crypto/keypair"
 	"github.com/ontio/ontology/common"
 )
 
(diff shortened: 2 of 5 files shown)


account/file_store.go
```diff
@@ -22,11 +22,15 @@ import (
...
 	"encoding/json"
+	"errors"
+	"fmt"
+
 	"github.com/ontio/ontology-crypto/keypair"
 	//"github.com/ontio/ontology/core/types"
-	"github.com/ontio/ontology/common"
 	"io/ioutil"
 	"os"
+
```
Honglei-Cong: go format imports ??

account/file_store.go
```diff
@@ -151,5 +172,48 @@ func (this *WalletData) Load(path string) error {
...
+func (this *WalletData) ToDefaultSecurity(passwords [][]byte) error {
+	return this.reencrypt(passwords, nil)
+}
+
+func (this *WalletData) reencrypt(passwords [][]byte, param *keypair.ScryptParam) error {
+	if len(passwords) != len(this.Accounts) {
+		return errors.New("not enough passwords for the accounts")
+	}
+	keys := make([]*keypair.ProtectedKey, len(this.Accounts))
+	for i, v := range this.Accounts {
```
Honglei-Cong: it seems the following two loops can merged together?

account/file_store.go
```diff
@@ -91,6 +95,23 @@ func NewWalletData() *WalletData {
 	}
 }
 
+func (this *WalletData) Dump() *WalletData {
```
Honglei-Cong, line 96: question: is this Copy() ?

cmd/account_cmd.go
```diff
@@ -424,3 +435,43 @@ func accountImport(ctx *cli.Context) error {
...
+				pwd, err := password.GetPassword()
+				if err != nil {
+					fmt.Println(err)
+				} else {
+					passwords[i] = pwd
+					break
+				}
+			}
+		}
+		wallet = wallet.Dump()
```
Honglei-Cong: this copied walletData to new instance, should client.WalletData be updated?

cmd/account_cmd.go
```diff
@@ -424,3 +435,43 @@ func accountImport(ctx *cli.Context) error {
...
+		err := wallet.ToLowSecurity(passwords)
+		if err != nil {
+			return fmt.Errorf("export failed: %s", err)
+		}
+	}
```
Honglei-Cong: if not to low security, to default security? or to original security?

account/client.go
```diff
@@ -313,7 +315,9 @@ func (this *ClientImpl) getAccount(accData *AccountData, passwd []byte) (*Accoun
 	if !accData.VerifyPassword(passwd) {
 		return nil, fmt.Errorf("verify password failed")
 	}
+	keypair.SetScryptParam(this.walletData.Scrypt)
 	privateKey, err := keypair.DecryptPrivateKey(&accData.ProtectedKey, passwd)
+	keypair.SetScryptParam(nil)
```
Honglei-Cong, line 320: question: changed to default security params here?

account/file_store.go
AlverLyu: Yes, deep copy.

cmd/account_cmd.go
AlverLyu: Same as the original wallet.

account/client.go
AlverLyu: Every encryption/decryption operation set the parameters as needed, and reset to default afterwards.

account/file_store.go
laizy: Unfortunately，it is not deep copy.  value returned by v.GetKeyPair will share internal data with the original one. and `Dump` is usually used to print some debug message.  `Clone` or `Copy` is a better one?

laizy: if you do need deep copy , you can use  `json.Marshal` and `json.UnMarshal`

AlverLyu: It does not matter actually. This function is used with reencrypt.

PR #199, opened 2018-05-21
[ONT-590] Cancel passwd input when prepare invoke contract

gasby88:

### DIFF
--- a/cmd/contract_cmd.go
+++ b/cmd/contract_cmd.go
@@ -177,14 +177,14 @@ func invokeContract(ctx *cli.Context) error {
 		}
 		return nil
 	}
-	singer, err := cmdcom.GetAccount(ctx)
+	signer, err := cmdcom.GetAccount(ctx)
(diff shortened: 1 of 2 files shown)


cmd/contract_cmd.go
```diff
@@ -183,6 +177,13 @@ func invokeContract(ctx *cli.Context) error {
 		}
 		return nil
 	}
+	singer, err := cmdcom.GetAccount(ctx)
```
laizy: singer -> signer

cmd/contract_cmd.go
```diff
@@ -183,6 +177,13 @@ func invokeContract(ctx *cli.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("Get singer account error:%s", err)
```
laizy: ditto

laizy (changes requested):

laizy (approved):

PR #207, opened 2018-05-22
ONT-604 refactor the auth contract

kunxian-xia:
The old contract has the problem that use a lot of storage to store a big table, i.e. OntID x Func -> authToken. In this PR, we make a tradeoff between space and time complexity by computing the table implicitly during runtime. In particular, we compute the table implicitly by join the OntID -> [](role, authToken) table and role -> []funcs.

### DIFF
--- a/smartcontract/service/native/auth/utils.go
+++ b/smartcontract/service/native/auth/utils.go
@@ -24,45 +24,186 @@ import (
 
 	"github.com/ontio/ontology/common/serialization"
 	"github.com/ontio/ontology/smartcontract/service/native"
+	"github.com/ontio/ontology/smartcontract/service/native/utils"
 )
 
 var (
-	RoleF        = []byte{0x01}
-	RoleP        = []byte{0x02}
-	DelegateList = []byte{0x04}
-	Admin        = []byte{0x05}
+	PreRoleFunc       = []byte{0x01}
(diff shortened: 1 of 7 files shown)


smartcontract/service/native/auth/utils.go
```diff
@@ -19,68 +19,224 @@
...
+	//PreDelegateList   = []byte{0x04}
 )
 
-//this.contractAddr.Admin
-func GetContractAdminKey(native *native.NativeService, contractAddr []byte) ([]byte, error) {
+//type(this.contractAddr.Admin) = []byte
+func concatContractAdminKey(native *native.NativeService, contractAddr []byte) ([]byte, error) {
 	this := native.ContextRef.CurrentContext().ContractAddress
-	adminKey, err := PackKeys(this[:], [][]byte{contractAddr, Admin})
+	adminKey, err := packKeys(this[:], [][]byte{contractAddr, PreAdmin})
```
laizy: `packKeys` serialize address in varbytes form, which may cause problem.

laizy (changes requested):

smartcontract/service/native/auth/utils.go
kunxian-xia: packKeys won't serialize the field param

laizy (approved):

PR #212, opened 2018-05-24
handle serialize err in block/header/vote_state

laizy:


core/states/vote_state.go
```diff
@@ -33,7 +33,10 @@ type VoteState struct {
 }
 
 func (this *VoteState) Serialize(w io.Writer) error {
-	this.StateBase.Serialize(w)
+	err := this.StateBase.Serialize(w)
+	if err != nil {
+		return err
+	}
 	serialization.WriteUint32(w, uint32(len(this.PublicKeys)))
```
gasby88: Will return error

gasby88 (approved):

PR #213, opened 2018-05-24
ONT-599 reconnect recent peers to avoid seeds service restart issue

BooniesFX:
Signed-off-by: Xiang Fu <fuxiang@onchain.com>

### DIFF
--- a/p2pserver/common/p2p_common.go
+++ b/p2pserver/common/p2p_common.go
@@ -76,6 +76,13 @@ const (
 	ACTOR_TIMEOUT = 5 //actor request timeout in secs
 )
 
+//recent contact const
+const (
+	RECENT_TIMEOUT   = 60
+	RECENT_FILE_NAME = "peers.recent"
+	RECENT_LIMIT     = 10 //recent contact list limit
+)
+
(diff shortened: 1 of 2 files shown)


p2pserver/common/p2p_common.go
```diff
@@ -76,6 +76,13 @@ const (
 	ACTOR_TIMEOUT = 5 //actor request timeout in secs
 )
 
+//recent contact const
+const (
+	RECENT_TIMEOUT   = 60
+	RECENT_FILE_NAME = "peers.recent"
```
laizy, line 82: how would it be if node run in mainnet -> restart -> testnet ? may be info should be stored or use different magic number in net message header .

BooniesFX: In case node would switch between mainnet  and testnet, the magic number should store inmsg header, and i think we should add net magic number in config which user can change it manually.
I will create a new issue about this situation, if check magic number when in hand shake state，node shouldn't`t connect to wrong peers and the recent list will be changed when valid connection setup.

laizy (approved):

PR #225, opened 2018-05-25
modify address scheme in line with neo

laizy:
work in progress. modification of witness in transaction will be submitted soon.


Arbio5zt (approved):
approve.

PR #227, opened 2018-05-25
change vrf to crypto.vrf

Honglei-Cong:


Honglei-Cong:
updated.
added key format checking when new peer join network, because only limited kinds of private keys can be used for VRF computation.

Honglei-Cong:
Governance native contract also need check if peer public key can support VRF computation.
This will be done by Menghang.

laizy (approved):

PR #246, opened 2018-05-31
ONT-682 add message payload size limit to avoid deserialize attack

chenzhijie:

### DIFF
--- a/p2pserver/common/p2p_common.go
+++ b/p2pserver/common/p2p_common.go
@@ -30,15 +30,15 @@ const (
 
 //msg cmd const
 const (
-	MSG_CMD_LEN      = 12  //msg type length in byte
-	CMD_OFFSET       = 4   //cmd type offet in msg hdr
-	CHECKSUM_LEN     = 4   //checksum length in byte
-	HASH_LEN         = 32  // hash length in byte
-	MSG_HDR_LEN      = 24  //msg hdr length in byte
-	MAX_BLK_HDR_CNT  = 500 //hdr count once when sync header
-	MAX_INV_HDR_CNT  = 500 //inventory count once when req inv
-	MAX_REQ_BLK_ONCE = 16  //req blk count once from one peer when sync blk
-
+	MSG_CMD_LEN      = 12               //msg type length in byte
+	CMD_OFFSET       = 4                //cmd type offet in msg hdr
+	CHECKSUM_LEN     = 4                //checksum length in byte
+	HASH_LEN         = 32               // hash length in byte
+	MSG_HDR_LEN      = 24               //msg hdr length in byte
+	MAX_BLK_HDR_CNT  = 500              //hdr count once when sync header
+	MAX_INV_HDR_CNT  = 500              //inventory count once when req inv
+	MAX_REQ_BLK_ONCE = 16               //req blk count once from one peer when sync blk
+	MAX_MSG_LEN      = 30 * 1024 * 1024 // the maximum message length
 )
 
 //info update const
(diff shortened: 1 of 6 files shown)


p2pserver/link/link.go
```diff
@@ -144,8 +144,14 @@ func unpackNodeBuf(this *Link, buf []byte) {
 			this.CloseConn()
 			return
 		}
-
-		rxBuf.len = types.PayloadLen(rxBuf.p)
+		payloadLen, err := types.PayloadLen(rxBuf.p)
+		if err != nil {
+			rxBuf.p = nil
+			rxBuf.len = 0
+			log.Warn("get payload len error")
```
laizy, line 151: need close the connection here since the connection is broken.

p2pserver/link/link_test.go
```diff
@@ -103,3 +111,113 @@ func TestNewLink(t *testing.T) {
...
+
+		blk.Transactions = txs
+		mBlk.Blk = blk
+
+		buf, err = mBlk.Serialization()
+		assert.Nil(t, err)
+	}
+
+	log.Infof("buf size: %d, %dM", len(buf), len(buf)/1024/1024)
+	unpackNodeBuf(cliLink, buf)
```
laizy: need assert for automated testing. rather than checking the execution log.

laizy (changes requested):

p2pserver/link/link_test.go
laizy: i mean assert the effects of `unpackNodeBuf`'s execution, clilink.rxBuf.len ==0 , conn is closed ect. or else this test function meaningless.

p2pserver/link/link.go
```diff
@@ -144,8 +144,15 @@ func unpackNodeBuf(this *Link, buf []byte) {
...
+			this.CloseConn()
+			return
+		}
+		rxBuf.len = payloadLen
```
Honglei-Cong, line 155: why not check MAX_MSG_LEN here?
if payloadLen > max_msg_len, close the connection?

chenzhijie: if payloadLen > max_msg_len, the PayloadLen function will return a error.

Honglei-Cong: Great. 
Then is it still necessary to checking MAX_MSG_LEN in the deserialization of all kinds of payload?
The payload length will never larger than max_msg_len.

chenzhijie: In this case, the msg received from link will be restricted of length. And if somewhere other call the deserialization function, it should do the same logic. So the checking inside the deserialization is a protect.

Honglei-Cong: ok

p2pserver/common/p2p_common.go
```diff
@@ -30,15 +30,20 @@ const (
...
-	MAX_REQ_BLK_ONCE = 16  //req blk count once from one peer when sync blk
+	MSG_CMD_LEN      = 12                //msg type length in byte
+	CMD_OFFSET       = 4                 //cmd type offet in msg hdr
+	CHECKSUM_LEN     = 4                 //checksum length in byte
+	HASH_LEN         = 32                // hash length in byte
+	MSG_HDR_LEN      = 24                //msg hdr length in byte
+	MAX_BLK_HDR_CNT  = 500               //hdr count once when sync header
+	MAX_INV_HDR_CNT  = 500               //inventory count once when req inv
+	MAX_REQ_BLK_ONCE = 16                //req blk count once from one peer when sync blk
+	MAX_MSG_LEN      = 100 * 1024 * 1024 // the maximum message length
```
laizy: consider to use a smaller value? maybe 30M?

chenzhijie: If the maximum transaction count is 600,000, each transaction is 200B。 600000 * 200B = 114M。  30M may used for at most 150,000 transaction. is it ok?

laizy: have comfirmed with @luodanwg . 30M is large enough for mainnet. so be conservative with this.

laizy: the default maximum block transaction count is 60000 actually, 600000 is written by mistake.

chenzhijie: ok, the commit has updated. please review it, thanks

laizy (approved):

PR #247, opened 2018-05-31
[ONT-469, ONT-806] add unit test, modify some failed test case

qiluge:


common/compact_uint_test.go
```diff
@@ -23,8 +23,14 @@ import (
...
-	a := uint64(1200000)
+	a := uint64(120)
 	compactValue := SetCompactUint(a)
 	unCompactValue, _ := GetCompactUint(compactValue)
 	assert.Equal(t, unCompactValue, a)
+
+	a = uint64(12000000)
+	compactValue = SetCompactUint(a)
+	assert.Panics(t, func() {
+		GetCompactUint(compactValue)
```
laizy: what‘s  this?

laizy (changes requested):

common/compact_uint_test.go
qiluge: i don't know, the compact_utils.go is not used in project.

laizy: I mean GetCompactUint panic means it is buggy.

PR #266, opened 2018-06-05
fix some bug in auth contract and push more event msg out

kunxian-xia:


smartcontract/service/native/auth/auth.go
```diff
@@ -106,20 +109,26 @@ func transfer(native *native.NativeService, contractAddr, newAdminOntID []byte,
...
+	//deserialize param
 	param := new(TransferParam)
 	rd := bytes.NewReader(native.Input)
 	err := param.Deserialize(rd)
 	if err != nil {
 		return nil, err
 	}
+
+	//prepare event msg
+	contract := hex.EncodeToString(param.ContractAddr)
```
laizy: use addr.ToHexString

smartcontract/service/native/auth/auth.go
```diff
@@ -131,8 +140,14 @@ func AssignFuncsToRole(native *native.NativeService) ([]byte, error) {
 	if err := param.Deserialize(rd); err != nil {
 		return nil, fmt.Errorf("deserialize param failed, caused by %v", err)
 	}
+
+	//prepare event msg
+	contract := hex.EncodeToString(param.ContractAddr)
```
laizy: and here

smartcontract/service/native/auth/auth.go
```diff
@@ -76,7 +77,9 @@ func InitContractAdmin(native *native.NativeService) ([]byte, error) {
 	if !ret {
 		return utils.BYTE_FALSE, nil
 	}
-	pushEvent(native, []interface{}{"initContractAdmin", param.AdminOntID})
+
+	msg := []interface{}{"initContractAdmin", hex.EncodeToString(invokeAddr[:]), param.AdminOntID}
```
laizy, line 81: use`ToHexString()`

smartcontract/service/native/auth/auth.go
```diff
@@ -131,83 +142,79 @@ func AssignFuncsToRole(native *native.NativeService) ([]byte, error) {
...
 	if param.Role == nil {
-		invokeEvent(native, "assignFuncsToRole", false)
+		pushEvent(native, failState)
 		return utils.BYTE_FALSE, nil
 	}
 
 	//check the caller's permission
-	admin, err := getContractAdmin(native, param.ContractAddr)
+	contractAddr, _ := addressToBytes(&param.ContractAddr)
+	admin, err := getContractAdmin(native, contractAddr)
```
laizy, line 158: why define `addressToBytes` function instead of using `getContractAdmin(native, param.ContractAddr[:])` directly?

smartcontract/service/native/auth/auth.go
```diff
@@ -307,29 +336,31 @@ func getLevel(native *native.NativeService, contractAddr, ontID, role []byte) (u
  * then make changes to storage as follows:
  */
 func delegate(native *native.NativeService, contractAddr []byte, from []byte, to []byte,
-	role []byte, period uint32, level uint8, keyNo uint64) ([]byte, error) {
+	role []byte, period uint32, level uint8, keyNo uint64) (bool, error) {
```
laizy, line 339: the type of param `from` `to` are Address?

kunxian-xia: No, they are ONT ID

smartcontract/service/native/auth/auth.go
kunxian-xia: Although they are equivalent, I prefer to use the `addressToBytes` function because it does not depend how the address type is defined.

smartcontract/service/native/auth/auth.go
kunxian-xia: okay

PR #267, opened 2018-06-06
fix missing total supply store in contract init function

laizy:


tanZiWen:
approve

PR #270, opened 2018-06-06
implement parallel block txn deserialization

Honglei-Cong:
about 70% performance improved on big-block deserialization

just for performance test reference, don't need to merge.

### DIFF
--- a/core/types/transaction_deserializer.go
+++ b/core/types/transaction_deserializer.go
@@ -1,0 +1,148 @@
+/*
+ * Copyright (C) 2018 The ontology Authors
+ * This file is part of The ontology library.
+ *
+ * The ontology is free software: you can redistribute it and/or modify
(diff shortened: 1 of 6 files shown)


core/types/transaction_deserializer.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	// all tx buffer send to inC
+	close(inC)
+	// wait deserializer workers done
+	deserializerWg.Wait()
+	// all tx deserializer done, all deserialized tx have sent to outC
+	close(outC)
+	// wait summarizer done
+	summarizerWg.Wait()
+
```
Honglei-Cong, line 132: check errC again?

laizy:
looks good, but it changes the block structure. so need more evaluation and exploration of alternative  scheme.

PR #272, opened 2018-06-07
[ONT-731]clirpc add sig to invoke neovm contract by abi

gasby88:


cmd/utils/neovm_param_abi.go
```diff
@@ -0,0 +1,124 @@
...
+	return string(data)
+}
+
+func ParseNeovmParam(params []interface{}, paramsAbi []*abi.NeovmContractParamsAbi) ([]interface{}, error) {
+	if len(params) != len(paramsAbi) {
+		return nil, fmt.Errorf("abi param not match")
+	}
+	val := make([]interface{}, 0)
+	for i, param := range params {
+		rawParam, ok := param.(string)
```
laizy: seems only accept string type here, so why not define `params` as `[]string`

laizy (approved):

PR #273, opened 2018-06-07
ONT-717 Update ontid contract

AlverLyu:
* optimize removing function
* add ID format checking in register

### DIFF
--- a/smartcontract/service/native/ontid/utils.go
+++ b/smartcontract/service/native/ontid/utils.go
@@ -18,9 +18,13 @@
 package ontid
 
 import (
+	"bytes"
+	"crypto/sha256"
 	"encoding/hex"
 	"errors"
+	"math/big"
 
+	"github.com/itchyny/base58-go"
 	"github.com/ontio/ontology-crypto/keypair"
 	com "github.com/ontio/ontology/common"
 	"github.com/ontio/ontology/core/states"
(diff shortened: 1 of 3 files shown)


smartcontract/service/native/ontid/utils.go
```diff
@@ -68,6 +72,30 @@ func decodeID(data []byte) ([]byte, error) {
 	return data[1:], nil
 }
 
+func verifyID(id []byte) bool {
+	if string(id[0:8]) != "did:ont:" {
```
laizy: obviously this will cause index out of bound panic .

laizy (changes requested):

smartcontract/service/native/ontid/utils.go
AlverLyu: fixed

smartcontract/service/native/ontid/utils.go
```diff
@@ -68,6 +72,33 @@ func decodeID(data []byte) ([]byte, error) {
...
+	if err != nil {
+		return false
+	}
+	bi, ok := new(big.Int).SetString(string(buf), 10)
+	if !ok || bi == nil {
+		return false
+	}
+	buf = bi.Bytes()
+	pos := len(buf) - 4
+	data := buf[:pos]
```
laizy: obviously this will cause panic too. how can you guarantee `len(buf) >= 4`?

laizy (changes requested):

laizy (approved):

PR #275, opened 2018-06-07
remove tx attribute

laizy:


qiluge:
I set uuid as the attribute to generate different transaction for stress test. If I remove the attribute, what should I use to distinguish a tx from other ?

laizy:
Transaction Nonce is designed for this purpose

qiluge:
ok, thanks

Arbio5zt (approved):

PR #276, opened 2018-06-07
ONT-750 Use the max gas price between global paramm and cmd to filter a tr…

lightshine001:
…ansaction

### DIFF
--- a/txnpool/proc/txnpool_server.go
+++ b/txnpool/proc/txnpool_server.go
@@ -21,14 +21,23 @@
 package proc
 
 import (
+	"bytes"
+	"encoding/hex"
+	"fmt"
 	"sort"
+	"strconv"
 	"sync"
 
 	"github.com/ontio/ontology-eventbus/actor"
+	"github.com/ontio/ontology/cmd/utils"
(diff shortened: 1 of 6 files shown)


txnpool/proc/txnpool_server.go
```diff
@@ -432,8 +504,21 @@ func (s *TXPoolServer) getPendingTxs(byCount bool) []*tx.Transaction {
...
 
 // cleanTransactionList cleans the txs in the block from the ledger
-func (s *TXPoolServer) cleanTransactionList(txs []*tx.Transaction) error {
-	return s.txPool.CleanTransactionList(txs)
+func (s *TXPoolServer) cleanTransactionList(txs []*tx.Transaction) {
+	s.txPool.CleanTransactionList(txs)
+
+	// Check whether to update the gas price and remove txs below the
+	// threshold
+	if ledger.DefLedger.GetCurrentBlockHeight()%tc.UPDATE_FREQUENCY == 0 {
```
laizy: get height from the  block complete event , rather than query db here

laizy (changes requested):

laizy (approved):

PR #279, opened 2018-06-08
close connection when handling error network message

lightshine001:


BooniesFX (approved):

laizy:
this case is handled in #278

PR #293, opened 2018-06-10
[ONT-794] Native contract integration with NEO

gasby88:


Arbio5zt:
please change commits to ontio:master

PR #306, opened 2018-06-11
[ONT-787]Create data dir according networkid

gasby88:

### DIFF
--- a/common/constants/constants.go
+++ b/common/constants/constants.go
@@ -18,7 +18,10 @@
 
 package constants
 
-import "time"
+import (
+	"fmt"
+	"time"
+)
 
 // genesis constants
 var (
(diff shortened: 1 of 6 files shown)


common/constants/constants.go
```diff
@@ -67,3 +70,40 @@ var UNBOUND_DEADLINE = (func() uint32 {
...
+	NETWORK_ID_POLARIS_NET   = 2
+	NETWORK_ID_SOLO_NET      = 3
+	NETWORK_NAME_MAIN_NET    = "ontology"
+	NETWORK_NAME_POLARIS_NET = "polaris"
+	NETWORK_NAME_SOLO_NET    = "testmode"
+)
+
+var NETWORK_MAGIC = map[uint32]uint32{
+	NETWORK_ID_MAIN_NET:    0x74746e41, //Network main
+	NETWORK_ID_POLARIS_NET: 0x01015401, //Network polaris
```
laizy: `constants.go` only define system wide constant.  so only define `const NETWORK_MAGIC_MAIN_NET = 0x74746e41, const NETWORK_MAGIC_POLARIS_NET = 0x01015401` in this file. all other variable and function should be reside in other place to avoid explosion.

laizy (approved):

PR #343, opened 2018-06-13
fix native param test

laizy:


gasby88:
I have fix in rp-341. So you can, close it

PR #361, opened 2018-06-19
fix remove connecting list bug

chenzhijie:

### DIFF
--- a/p2pserver/net/netserver/netserver.go
+++ b/p2pserver/net/netserver/netserver.go
@@ -483,9 +483,9 @@ func (this *NetServer) RemoveFromConnectingList(addr string) {
 	this.ConnectingNodes.Lock()
 	defer this.ConnectingNodes.Unlock()
 	addrs := []string{}
-	for i, a := range this.ConnectingAddrs {
-		if strings.Compare(a, addr) == 0 {
-			addrs = append(this.ConnectingAddrs[:i], this.ConnectingAddrs[i+1:]...)
+	for _, a := range this.ConnectingAddrs {
+		if strings.Compare(a, addr) != 0 {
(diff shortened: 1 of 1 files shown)


p2pserver/net/netserver/netserver.go
```diff
@@ -483,9 +483,9 @@ func (this *NetServer) RemoveFromConnectingList(addr string) {
 	this.ConnectingNodes.Lock()
 	defer this.ConnectingNodes.Unlock()
 	addrs := []string{}
-	for i, a := range this.ConnectingAddrs {
-		if strings.Compare(a, addr) == 0 {
-			addrs = append(this.ConnectingAddrs[:i], this.ConnectingAddrs[i+1:]...)
+	for _, a := range this.ConnectingAddrs {
+		if strings.Compare(a, addr) != 0 {
+			addrs = append(addrs, a)
```
lightshine001: Why not use this.ConnectingAddrs = append(this.ConnectingAddrs[:i], this.ConnectingAddrs[i+1:]...), instead of using temp variable addrs ?

chenzhijie: Because if you change the this.ConnectingAddrs which is used for loop, it is dangerous. And in this case, it also can not remove duplicated elements.

lightshine001: Ok. So please use addrs := this.ConnectingAddrs[:0] to avoid allocating a new underlying array

laizy: btw: you can use ` if a != addr` directly

chenzhijie: That's right, thanks all

laizy (approved):

PR #363, opened 2018-06-19
Update cli doc

gasby88:

### DIFF
--- a/docs/specifications/cli_user_guide.md
+++ b/docs/specifications/cli_user_guide.md
@@ -13,7 +13,7 @@ The following are the command line parameters supported by Ontology CLI:
 #### 1.1.1 Ontology System Parameters
 
 --config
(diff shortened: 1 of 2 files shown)


docs/specifications/cli_user_guide.md
```diff
@@ -1,570 +1,803 @@
...
-```
-./ontology help
-./ontology <command> help
-```
----
+#### 1.1.1 Ontology System Parameters
 
-## <a name="cli-account-cmd"></a> CLI Account Command
+--config
+The config parameter specifies the file path of the genesis block for the current Ontolgy node. If don't specifies, Ontology will use the config of Polaris test net. Note that the genesis block configuration must be the same for all nodes in the same network, otherwise it will not be able to synchronize blocks or start nodes due to block data incompatibility.
```
laizy: `don't specifies` -> ` does't specify `

laizy (approved):

PR #364, opened 2018-06-19
ONT-892 modify Proposer peers number must great than or equal to chain.C

xiemylogos:


consensus/vbft/node_utils.go
```diff
@@ -209,7 +209,7 @@ func (self *Server) buildParticipantConfig(blkNum uint32, block *Block, chainCfg
 	}
 
 	s := 0
-	cfg.Proposers = calcParticipantPeers(cfg, chainCfg, s, s+vconfig.MAX_PROPOSER_COUNT)
+	cfg.Proposers = calcParticipantPeers(cfg, chainCfg, s, s+vconfig.MAX_ENDORSER_COUNT)
```
laizy: why use `MAX_ENDORSER_COUNT` here?

laizy (approved):

PR #391, opened 2018-06-21
ONT-924 optimize transmit logic and the compatibility of dual/single port

BooniesFX:


p2pserver/link/link.go
```diff
@@ -160,9 +160,25 @@ func (this *Link) Tx(msg types.Message) error {
...
 
-	_, err = conn.Write(buf.Bytes())
+	payload := buf.Bytes()
+	nByteCnt := len(payload)
+	log.Debugf("TX buf length: %d\n", nByteCnt)
+
+	nCount := nByteCnt / common.PER_SEND_LEN
+	for i := 0; i < nCount; i++ {
+		conn.SetWriteDeadline(time.Now().Add(common.WRITE_DEADLINE * time.Second))
+		_, err = conn.Write(payload[i*common.PER_SEND_LEN : (i+1)*common.PER_SEND_LEN])
```
laizy: any reason why need divided to n parts to call conn.Write instead one ?

BooniesFX: small package would be transmit more faster than write whole buffer under deadline and poor network environment

laizy: It's the exact opposite of what I think.

laizy (approved):

PR #405, opened 2018-06-22
add the limitation of serialize stackitem

JasonZhouPW:


smartcontract/service/neovm/runtime.go
```diff
@@ -385,3 +432,28 @@ func circularRefDetection(value vmtypes.StackItems, visited map[uintptr]bool, de
 
 	return false
 }
+
+func checkWroteBytesLen(length int) bool {
+	if length > VM_SERIALIZE_MAX_LEN {
```
Honglei-Cong, line 437: just return length > VM_SERIALIZE_MAX_LEN ?

PR #406, opened 2018-06-22
ONT-897 add ontID format check in several functions

kunxian-xia:
- initContractAdmin
- transfer
- assignOntIDsToRole
- delegate

### DIFF
--- a/smartcontract/service/native/auth/auth.go
+++ b/smartcontract/service/native/auth/auth.go
@@ -28,6 +28,7 @@ import (
 	"github.com/ontio/ontology/common/serialization"
 	"github.com/ontio/ontology/errors"
 	"github.com/ontio/ontology/smartcontract/service/native"
+	"github.com/ontio/ontology/smartcontract/service/native/ontid"
 	"github.com/ontio/ontology/smartcontract/service/native/utils"
 )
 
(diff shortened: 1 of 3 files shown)


smartcontract/service/native/auth/auth.go
```diff
@@ -70,6 +71,9 @@ func InitContractAdmin(native *native.NativeService) ([]byte, error) {
 	}
 	invokeAddr := cxt.ContractAddress
 
+	if !ontid.VerifyID(param.AdminOntID) {
+		return utils.BYTE_FALSE, nil
```
Honglei-Cong: do we need provide some error information?

kunxian-xia: push error out?

Honglei-Cong: just a question :)

return BYTE_FALSE, any error-info?

kunxian-xia: In this case, the param provided is not valid, so I choose to return false for there is no internal-error happened. Maybe I should add some error-info log?

smartcontract/service/native/ontid/utils.go
```diff
@@ -72,7 +72,7 @@ func decodeID(data []byte) ([]byte, error) {
 	return data[1:], nil
 }
 
-func verifyID(id []byte) bool {
+func VerifyID(id []byte) bool {
```
laizy: remove this duplicated code and use the one in `account/identity.go`

kunxian-xia: already done : )

laizy (approved):

PR #427, opened 2018-06-26
modify stackitem equals

JasonZhouPW:


laizy:
asymmetry implementation may cause `true == 10` but `10 != true`

smartcontract/test/panic_test.go
```diff
@@ -65,7 +65,7 @@ func TestRandomCodeCrash(t *testing.T) {
 			//cache := storage.NewCloneCache(testBatch)
 			sc := SmartContract{
 				Config:     config,
-				Gas:        10000,
+				Gas:        0,
```
laizy: why modify this?

vm/neovm/types/struct.go
```diff
@@ -41,7 +41,17 @@ func NewStruct(value []StackItems) *Struct {
...
+	if this == other {
+		return true
+	}
+
+	oa, err := other.GetStruct()
+	if err != nil {
+		return false
+	}
+
+	return reflect.DeepEqual(this._array, oa)
```
laizy, line 54: this is unsound. `Equals` on `ByteArray, Integer and Bool ` now is type-erased, but  `DeepEqual` is not.

tanZiWen (approved):

PR #430, opened 2018-06-26
return nil if tx not in txpool

xizho10:


http/base/rest/interfaces.go
```diff
@@ -529,7 +529,8 @@ func GetMemPoolTxState(cmd map[string]interface{}) map[string]interface{} {
 	}
 	txEntry, err := bactor.GetTxFromPool(hash)
 	if err != nil {
-		return ResponsePack(berr.UNKNOWN_TRANSACTION)
+		resp["Result"] = nil
+		return resp
```
laizy, line 533: why not report error?

PR #450, opened 2018-06-28
[ONT-975] Optimize Ontology consensus parameter

gasby88:
1、Wont't turn consensus module on by default
2、Wont't need wallet when startup a synchronization node


main.go
```diff
@@ -207,6 +207,9 @@ func initConfig(ctx *cli.Context) (*config.OntologyConfig, error) {
 }
 
 func initAccount(ctx *cli.Context) (*account.Account, error) {
+	if !config.DefConfig.Consensus.EnableConsensus {
+		return nil, nil
```
laizy, line 211: fix panic `	log.Infof("Using account:%s", acc.Address.ToBase58())` in main.go since you return nil here

laizy (approved):

PR #483, opened 2018-06-30
Fix the colon format in log and add more info in HeadersReqHandle

BooniesFX:


p2pserver/message/utils/msg_handler.go
```diff
@@ -87,12 +87,12 @@ func HeadersReqHandle(data *msgTypes.MsgPayload, p2p p2p.P2P, pid *evtActor.PID,
 
 	headers, err := GetHeadersFromHash(startHash, stopHash)
 	if err != nil {
-		log.Error(err)
+		log.Errorf("get headers in HeadersReqHandle error: %s,startHash:%x,stopHash:%x", err.Error(), startHash, stopHash)
```
laizy: the hex string for hash is reversed , so use `hash.ToHexString` instead.

consensus/vbft/msg_builder.go
```diff
@@ -201,7 +201,7 @@ func (self *Server) constructBlock(blkNum uint32, prevBlkHash common.Uint256, tx
 	blkHash := blk.Hash()
 	sig, err := signature.Sign(self.account, blkHash[:])
 	if err != nil {
-		return nil, fmt.Errorf("sign block failed, block hash：%x, error: %s", blkHash, err)
+		return nil, fmt.Errorf("sign block failed, block hash:%x, error: %s", blkHash, err)
```
laizy: and this

Arbio5zt (approved):

PR #514, opened 2018-07-05
Update README.md

panther142:
- Add notes for when networkid is a custom value
- Add notes for enabling consensus nodes
- Update some examples to show Mainnet address instead of Testnet addresses for consistency with the rest of the document

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,8 @@ Here's a example of single-host configuration:
 
 #### Multi-hosts deployment configuration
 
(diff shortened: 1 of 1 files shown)


qiluge:
Thank you very much，would you like to update Chinese version at the same time?

panther142:
Unfortunately i can't as I am not a chinese speaker. But if someone from your team could that would be great.

Arbio5zt (approved):

PR #517, opened 2018-07-05
ONT-1017：add PreExecCheck in interface

xizho10:
1. add PreExecCheck in send transaction
2. add height in transaction json data 
3. add comment in interface for godoc

### DIFF
--- a/http/base/common/common.go
+++ b/http/base/common/common.go
@@ -16,6 +16,7 @@
  * along with The ontology.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+// Package common privides functions for http handler call
 package common
 
 import (
--- a/http/base/common/common.go
+++ b/http/base/common/common.go
@@ -24,17 +24,21 @@ import (
 	"fmt"
 	"github.com/ontio/ontology-crypto/keypair"
 	"github.com/ontio/ontology/common"
+	"github.com/ontio/ontology/common/config"
 	"github.com/ontio/ontology/common/log"
+	"github.com/ontio/ontology/core/ledger"
 	"github.com/ontio/ontology/core/payload"
 	"github.com/ontio/ontology/core/types"
 	ontErrors "github.com/ontio/ontology/errors"
 	bactor "github.com/ontio/ontology/http/base/actor"
 	"github.com/ontio/ontology/smartcontract/event"
+	"github.com/ontio/ontology/smartcontract/service/native/global_params"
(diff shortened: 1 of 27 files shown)


http/base/common/common.go
```diff
@@ -521,3 +527,88 @@ func BuildNeoVMParam(builder *neovm.ParamsBuilder, smartContractParams []interfa
...
+	}
+	result, err := ledger.DefLedger.PreExecuteContract(txn)
+	if err != nil {
+		return fmt.Errorf("PreExecuteContract fail")
+	}
+	if txn.GasLimit < result.Gas {
+		return fmt.Errorf("GasLimit should >= %d", result.Gas)
+	}
+	gas, overflow := common.SafeMul(txn.GasPrice, result.Gas)
+	if overflow {
```
laizy, line 556: have checked in line 523.

http/base/common/common.go
```diff
@@ -521,3 +527,88 @@ func BuildNeoVMParam(builder *neovm.ParamsBuilder, smartContractParams []interfa
...
+	if _, overflow := common.SafeMul(txn.GasLimit, txn.GasPrice); overflow {
+		return fmt.Errorf("Tx GasPrice %d and gasLimit %d safeMul overflow",
+			txn.GasPrice, txn.GasLimit)
+	}
+	if txn.TxType == types.Deploy && txn.GasLimit < svrneovm.CONTRACT_CREATE_GAS {
+		return fmt.Errorf("Deploy tx gasLimit should >= %d",
+			svrneovm.CONTRACT_CREATE_GAS)
+	}
+	gasLimitConfig := config.DefConfig.Common.GasLimit
+	gasPriceConfig := getGasPriceConfig()
```
laizy, line 543: i think you need cache this value and  refresh it  when it is expired, instead of fetching it every time.

laizy: given consensus node will disable pre-execute, performance is not crucial for client node, so keep it like this.

http/base/common/common.go
xizho10: The two check gasLimit is different

laizy: result.Gas <= txn.GasLimit,  so if line 523 passed, it will guarantee not overflow

xizho10: line 523  for get gas, not only for check overflow.

laizy (approved):

PR #521, opened 2018-07-06
add more debug info in p2p

BooniesFX:
1.add [p2p]prefix to make filtering msg easier
2.add more info in debug, include peer address and msg package
3.fix some legacy test case

### DIFF
--- a/p2pserver/actor/server/actor.go
+++ b/p2pserver/actor/server/actor.go
@@ -51,15 +51,15 @@ func (this *P2PActor) Start() (*actor.PID, error) {
 func (this *P2PActor) Receive(ctx actor.Context) {
 	switch msg := ctx.Message().(type) {
 	case *actor.Restarting:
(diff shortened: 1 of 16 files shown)


p2pserver/actor/server/actor.go
```diff
@@ -51,15 +51,15 @@ func (this *P2PActor) Start() (*actor.PID, error) {
 func (this *P2PActor) Receive(ctx actor.Context) {
 	switch msg := ctx.Message().(type) {
 	case *actor.Restarting:
-		log.Info("p2p actor restarting")
+		log.Debug("[p2p]actor restarting")
```
laizy: `restarting` means panic encountered, so debug level is not enough

BooniesFX: change log to Warn except the start state

laizy (approved):

PR #525, opened 2018-07-08
Optimize log print

lightshine001:


laizy (dismissed):

smartcontract/service/neovm/neovm_service.go
```diff
@@ -255,7 +255,7 @@ func (this *NeoVmService) getContract(address []byte) ([]byte, error) {
 	if err != nil {
 		return nil, errors.NewErr("[getContract] Get contract context error!")
 	}
-	log.Infof("invoke contract address:%x", scommon.ToArrayReverse(address))
+	log.Debugf("invoke contract address:%x", scommon.ToArrayReverse(address))
```
laizy, line 258: use address.ToHexString

txnpool/proc/txnpool_server.go
```diff
@@ -172,6 +170,7 @@ func (s *TXPoolServer) init(num uint8, preExec bool) {
 	}
 
 	s.gasPrice = getGasPriceConfig()
+	log.Infof("tx pool: initial enforced gas price %d", s.gasPrice)
```
Arbio5zt: I think this message be " The current local gasprice is :xxxx " would be better.

smartcontract/service/neovm/neovm_service.go
lightshine001: []byte has no method ToHexString

txnpool/proc/txnpool_server.go
lightshine001: Updated

laizy (approved):

PR #532, opened 2018-07-12
try specific recent peers by network id

BooniesFX:
check configuration before try recent peer
lower some network activities log level

### DIFF
--- a/p2pserver/p2pserver.go
+++ b/p2pserver/p2pserver.go
@@ -54,7 +54,7 @@ type P2PServer struct {
 	blockSync *BlockSyncMgr
 	ledger    *ledger.Ledger
 	ReconnectAddrs
-	recentPeers    []string
+	recentPeers    map[uint32][]string
 	quitSyncRecent chan bool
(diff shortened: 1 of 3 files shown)


laizy:
use `map[magicnumber][]Peer` like structure in your peer.recent file, rather than specialized case handle in code.

BooniesFX:
@laizy the file mainly used in main net, and the customize net don`t need it.

p2pserver/p2pserver.go
```diff
@@ -77,7 +77,7 @@ func NewServer() *P2PServer {
 
 	p.msgRouter = utils.NewMsgRouter(p.network)
 	p.blockSync = NewBlockSyncMgr(p)
-	p.recentPeers = make([]string, common.RECENT_LIMIT)
+	p.recentPeers = make(map[uint32][]string, common.RECENT_LIMIT)
```
laizy: `RECENT_LIMIT` limits the size of array, not the map.

laizy (approved):

PR #554, opened 2018-07-20
ONT-1091 optimize block sync

chenzhijie:
1. record peer's network status, timeout/error times
2. remove the error node from the list
3. request more nodes for min height block
4. combine same get data request 
5. add block cache to avoid ledger operation frequently
6. adjust send ping message period

### DIFF
--- a/p2pserver/block_sync.go
+++ b/p2pserver/block_sync.go
@@ -19,7 +19,6 @@
 package p2pserver
 
 import (
-	"bytes"
 	"math"
 	"sort"
 	"sync"
@@ -44,6 +43,7 @@ const (
 	SYNC_NEXT_BLOCK_TIMES        = 3          //Request times of next height block
 	SYNC_NEXT_BLOCKS_HEIGHT      = 2          //for current block height plus next
 	SYNC_NODE_RECORD_SPEED_CNT   = 3          //Record speed count for accuracy
+	SYNC_NODE_RECORD_TIME_CNT    = 3          //Record request time  for accuracy
 	SYNC_NODE_SPEED_INIT         = 100 * 1024 //Init a big speed (100MB/s) for every node in first round
 	SYNC_MAX_ERROR_RESP_TIMES    = 5          //Max error headers/blocks response times, if reaches, delete it
 )
@@ -51,10 +51,10 @@ const (
 //NodeWeight record some params of node, using for sort
 type NodeWeight struct {
 	id           uint64    //NodeID
-	speed        []float32 //Record node request-response speed, using for calc the avg spped, unit kB/s
+	speed        []float32 //Record node request-response speed, using for calc the avg speed, unit kB/s
 	timeoutCnt   int       //Node response timeout count
 	errorRespCnt int       //Node response error data count
-	reqTime      int64     //Last request time
+	reqTime      []int64   //Record request time, using for calc the avg req time interval, unit millisecond
 }
 
 //NewNodeWeight new a nodeweight
@@ -63,12 +63,17 @@ func NewNodeWeight(id uint64) *NodeWeight {
 	for i := 0; i < SYNC_NODE_RECORD_SPEED_CNT; i++ {
 		s = append(s, float32(SYNC_NODE_SPEED_INIT))
 	}
+	r := make([]int64, 0, SYNC_NODE_RECORD_TIME_CNT)
+	now := time.Now().UnixNano() / int64(time.Millisecond)
+	for i := 0; i < SYNC_NODE_RECORD_TIME_CNT; i++ {
+		r = append(r, now)
+	}
 	return &NodeWeight{
 		id:           id,
 		speed:        s,
 		timeoutCnt:   0,
 		errorRespCnt: 0,
-		reqTime:      time.Now().UnixNano() / int64(time.Millisecond),
+		reqTime:      r,
 	}
 }
 
@@ -87,9 +92,10 @@ func (this *NodeWeight) GetErrorRespCnt() int {
 	return this.errorRespCnt
 }
 
-//ResetReqTime reset request time
-func (this *NodeWeight) ResetReqTime() {
-	this.reqTime = time.Now().UnixNano() / int64(time.Millisecond)
+//AppendNewReqTime append new request time
+func (this *NodeWeight) AppendNewReqtime() {
+	copy(this.reqTime[0:SYNC_NODE_RECORD_TIME_CNT-1], this.reqTime[1:])
+	this.reqTime[SYNC_NODE_RECORD_TIME_CNT-1] = time.Now().UnixNano() / int64(time.Millisecond)
 }
 
 //addNewSpeed apend the new speed to tail, remove the oldest one
@@ -100,16 +106,19 @@ func (this *NodeWeight) AppendNewSpeed(s float32) {
 
 //Weight calculate node's weight for sort. Highest weight node will be accessed first for next request.
 func (this *NodeWeight) Weight() float32 {
-	avg := float32(0.0)
+	avgSpeed := float32(0.0)
 	for _, s := range this.speed {
-		avg += s
+		avgSpeed += s
 	}
-	avg = avg / float32(len(this.speed))
-	var reqInterval int64
-	if this.reqTime != 0 {
-		reqInterval = time.Now().UnixNano()/int64(time.Millisecond) - this.reqTime
+	avgSpeed = avgSpeed / float32(len(this.speed))
+
+	avgInterval := float32(0.0)
+	now := time.Now().UnixNano() / int64(time.Millisecond)
+	for _, t := range this.reqTime {
+		avgInterval += float32(now - t)
 	}
-	w := avg + float32(reqInterval) - float32(this.timeoutCnt+this.errorRespCnt)
+	avgInterval = avgInterval / float32(len(this.reqTime))
+	w := avgSpeed + avgInterval
 	return w
 }
 
@@ -127,7 +136,9 @@ func (nws NodeWeights) Less(i, j int) bool {
 	if nws[i] == nil || nws[j] == nil {
 		return false
 	}
-	return nws[i].Weight() < nws[j].Weight()
+	ni := nws[i]
+	nj := nws[j]
+	return ni.Weight() < nj.Weight() && ni.errorRespCnt >= nj.errorRespCnt && ni.timeoutCnt >= nj.timeoutCnt
 }
 
 //SyncFlightInfo record the info of fight object(header or block)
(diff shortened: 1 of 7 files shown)


p2pserver/link/link.go
```diff
@@ -22,27 +22,34 @@ import (
...
+	respCache *lru.ARCCache          //Cache used for save response data, avoid legder operation
 }
 
 func NewLink() *Link {
-	link := &Link{}
-
+	link := &Link{
+		reqRecord: make(map[string]int64, 0),
+	}
+	link.respCache, _ = lru.NewARC(common.MAX_RESP_CACHE_SIZE)
```
BooniesFX: I think setup cache in handler would more useful than in link

p2pserver/block_sync.go
```diff
@@ -117,30 +206,34 @@ func (this *SyncFlightInfo) GetStartTime() time.Time {
...
+	nodeList       []uint64                             //Holder all of nodes that can be used
+	nextNodeIndex  int                                  //Index for polling node
+	server         *P2PServer                           //Pointer to the local node
+	syncBlockLock  bool                                 //Help to avoid send block sync request duplicate
+	syncHeaderLock bool                                 //Help to avoid send header sync request duplicate
+	saveBlockLock  bool                                 //Help to avoid saving block concurrently
+	exitCh         chan interface{}                     //ExitCh to receive exit signal
+	ledger         *ledger.Ledger                       //ledger
+	lock           sync.RWMutex                         //lock
+	recvBlkNodes   map[common.Uint256]uint64            //Map received BlockHash => NodeID, using for track wrong node
```
laizy: the values in this map are never been cleared, which will cause memory attack. Since the NodeId only used in save block logic, define blocksCache to something like `map[uint32] struct { NodeID, *types.Block}`

p2pserver/block_sync.go
```diff
@@ -33,14 +35,101 @@ import (
...
+	SYNC_MAX_ERROR_RESP_TIMES    = 5          //Max error headers/blocks response times, if reaches, delete it
 )
 
+//NodeWeight record some params of node, using for sort
+type NodeWeight struct {
+	id           uint64    //NodeID
+	speed        []float32 //Record node request-response speed, using for calc the avg spped, unit kB/s
+	timeoutCnt   int       //Node response timeout count
+	errorRespCnt int       //Node response error data count
+	reqTime      int64     //Last request time
```
laizy: multiple requests will send to a single node,  `reqTime` will override the time previous request has set.

p2pserver/block_sync.go
```diff
@@ -33,14 +35,101 @@ import (
...
+	SYNC_NEXT_BLOCKS_HEIGHT      = 2          //for current block height plus next
+	SYNC_NODE_RECORD_SPEED_CNT   = 3          //Record speed count for accuracy
+	SYNC_NODE_SPEED_INIT         = 100 * 1024 //Init a big speed (100MB/s) for every node in first round
+	SYNC_MAX_ERROR_RESP_TIMES    = 5          //Max error headers/blocks response times, if reaches, delete it
 )
 
+//NodeWeight record some params of node, using for sort
+type NodeWeight struct {
+	id           uint64    //NodeID
+	speed        []float32 //Record node request-response speed, using for calc the avg spped, unit kB/s
```
laizy: spped ->speed

p2pserver/block_sync.go
```diff
@@ -345,17 +464,27 @@ func (this *BlockSyncMgr) OnHeaderReceive(headers []*types.Header) {
...
+func (this *BlockSyncMgr) OnBlockReceive(fromID uint64, block *types.Block) {
 	height := block.Header.Height
 	blockHash := block.Hash()
 	log.Debugf("OnBlockReceive Height:%d", height)
+	this.recvBlkNodes[blockHash] = fromID
+	flightInfos := this.flightBlocks[blockHash]
+	for _, flightInfo := range flightInfos {
+		if flightInfo.GetNodeId() == fromID {
+			t := flightInfo.GetStartTime()
+			go this.addNewSpeed(fromID, block, t)
```
laizy: why need `go ` here?

p2pserver/block_sync.go
```diff
@@ -590,19 +736,28 @@ func (this *BlockSyncMgr) delFlightBlock(blockHash common.Uint256) bool {
...
+	}
+	return cnt
 }
 
 func (this *BlockSyncMgr) isBlockOnFlight(blockHash common.Uint256) bool {
-	flightInfo := this.getFlightBlock(blockHash)
-	return flightInfo != nil
+	flightInfos := this.getFlightBlock(blockHash)
+	for _, flightInfo := range flightInfos {
+		if flightInfo != nil {
```
laizy: explain when `flightInfo` will be `nil`. I think `return len(flightInfos) != 0` should be ok.

p2pserver/block_sync.go
```diff
@@ -611,13 +766,23 @@ func (this *BlockSyncMgr) getNextNodeId() uint64 {
...
+	weights := make(NodeWeights, 0, len(this.nodeWeights))
+	for _, w := range this.nodeWeights {
+		weights = append(weights, w)
+	}
+	sort.Sort(sort.Reverse(weights))
+	nodelist := make([]uint64, 0)
+	for _, n := range weights {
+		nodelist = append(nodelist, n.id)
+	}
+	this.nextNodeIndex = 0
```
laizy: since you set nextNodeIndex to 0 every time. delete this field in BlockSyncMgr and make it as a local variable.

p2pserver/block_sync.go
```diff
@@ -669,3 +834,42 @@ func (this *BlockSyncMgr) getNodeWithMinFailedTimes(flightInfo *SyncFlightInfo,
...
+		ns.ResetReqTime()
+	}
+}
+
+//addNewSpeed apend the new speed to tail, remove the oldest one
+func (this *BlockSyncMgr) addNewSpeed(nodeId uint64, block *types.Block, reqTime time.Time) {
+	//calulate response speed, get block's size may use some mem/io resouce
+	t := time.Now().UnixNano()/int64(time.Millisecond) - reqTime.UnixNano()/int64(time.Millisecond)
+	buf := bytes.NewBuffer([]byte{})
+	err := block.Serialize(buf)
```
laizy: This is too expensive. serialize the whole block just to get it's size???

p2pserver/link/link.go
```diff
@@ -180,3 +189,41 @@ func (this *Link) Tx(msg types.Message) error {
...
+	if msg.CmdType() != common.GET_DATA_TYPE {
+		return
+	}
+	now := time.Now().Unix()
+	if len(this.reqRecord) >= common.MAX_REQ_RECORD_SIZE-1 {
+		newRecord := make(map[string]int64, 0)
+		for id := range this.reqRecord {
+			t := this.reqRecord[id]
+			if int(now-t) < common.REQ_INTERVAL {
+				newRecord[id] = t
```
laizy: clean it inplace,  instead of copying all the valid entries every time.

laizy (changes requested):

p2pserver/block_sync.go
chenzhijie: Because addNewSpeed need to serialize block and it is a expensive operation. And the network speed is decided by the block size and the time spent, it seems the serialization is unavoidable。

p2pserver/link/link.go
chenzhijie: ok

p2pserver/block_sync.go
chenzhijie: ok

p2pserver/block_sync.go
chenzhijie: ok. my mistake

p2pserver/block_sync.go
chenzhijie: ok

p2pserver/block_sync.go
chenzhijie: great, thanks

p2pserver/block_sync.go
laizy: modify `message.go:ReadMessage` function to return `(Message, PayloadSize, error)`, the PayloadSize is just `hdr.Length`,  and add a `PayloadSize` field in `types.MsgPayload` struct.

chenzhijie: the code has updated, thanks

p2pserver/block_sync.go
```diff
@@ -745,28 +762,26 @@ func (this *BlockSyncMgr) getFlightBlockCount() int {
...
+		return true
 	}
 	return false
 }
 
 //Using polling for load balance
-func (this *BlockSyncMgr) getNextNodeId(nodeList []uint64) uint64 {
+func (this *BlockSyncMgr) getNextNodeId(nextNodeIndex int, nodeList []uint64) (int, uint64) {
 	this.lock.Lock()
 	defer this.lock.Unlock()
```
laizy: this function has nothing to do with `BlockSyncMgr` now.  remove this lock, and make it a free function

p2pserver/block_sync.go
```diff
@@ -127,7 +136,9 @@ func (nws NodeWeights) Less(i, j int) bool {
 	if nws[i] == nil || nws[j] == nil {
```
laizy: explain when it will be nil.

chenzhijie: I think getNextNode() is not a concurrent-safe function, and it may access a deleting node. What do you think?

laizy: You need synchronize the access of  `nodeWeights`

chenzhijie: ok. I have modified it, and btw i add a `delNode` operation when a node reaches max error times now.

p2pserver/block_sync.go
```diff
@@ -115,32 +212,39 @@ func (this *SyncFlightInfo) GetStartTime() time.Time {
...
-	syncBlockLock  bool                               //Help to avoid send block sync request duplicate
-	syncHeaderLock bool                               //Help to avoid send header sync request duplicate
-	saveBlockLock  bool                               //Help to avoid saving block concurrently
-	exitCh         chan interface{}                   //ExitCh to receive exit signal
-	ledger         *ledger.Ledger
-	lock           sync.RWMutex
+	flightBlocks   map[common.Uint256][]*SyncFlightInfo //Map BlockHash => []SyncFlightInfo, using for manager all of those block flights
+	flightHeaders  map[uint32]*SyncFlightInfo           //Map HeaderHeight => SyncFlightInfo, using for manager all of those header flights
+	blocksCache    map[uint32]*BlockInfo                //Map BlockHash => BlockInfo, using for cache the blocks receive from net, and waiting for commit to ledger
+	nodeList       []uint64                             //Holder all of nodes that can be used
```
laizy: `nodeList` is useless now.

laizy (approved):

PR #560, opened 2018-07-24
optimize hash calculation of transaction and block header.

laizy:


laizy:
optimization is implemented in #615 , so close this.

PR #566, opened 2018-07-26
avoid data copy and buffer allocation when read out p2p message from network

laizy:

### DIFF
--- a/p2pserver/link/link.go
+++ b/p2pserver/link/link.go
@@ -25,7 +25,7 @@ import (
 	"net"
 	"time"
 
-	common2 "github.com/ontio/ontology/common"
+	comm "github.com/ontio/ontology/common"
 	"github.com/ontio/ontology/common/log"
 	"github.com/ontio/ontology/p2pserver/common"
 	"github.com/ontio/ontology/p2pserver/message/types"
@@ -167,7 +167,7 @@ func (this *Link) Tx(msg types.Message) error {
 		return errors.New("[p2p]tx link invalid")
 	}
 
-	sink := common2.NewZeroCopySink(nil)
+	sink := comm.NewZeroCopySink(nil)
 	err := types.WriteMessage(sink, msg)
 	if err != nil {
 		log.Debugf("[p2p]error serialize messge ", err.Error())
--- a/p2pserver/link/link.go
+++ b/p2pserver/link/link.go
@@ -20,12 +20,12 @@ package link
 
 import (
 	"bufio"
-	"bytes"
 	"errors"
 	"fmt"
 	"net"
 	"time"
 
+	common2 "github.com/ontio/ontology/common"
 	"github.com/ontio/ontology/common/log"
 	"github.com/ontio/ontology/p2pserver/common"
 	"github.com/ontio/ontology/p2pserver/message/types"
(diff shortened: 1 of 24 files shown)


lightshine001 (dismissed):
It’s ok for me. Please confirm ut passed.

p2pserver/link/link.go
```diff
@@ -20,12 +20,12 @@ package link
 
 import (
 	"bufio"
-	"bytes"
 	"errors"
 	"fmt"
 	"net"
 	"time"
 
+	common2 "github.com/ontio/ontology/common"
```
BooniesFX: use another name rather than "pkg+2"??

p2pserver/message/types/address.go
```diff
@@ -32,44 +33,70 @@ type Addr struct {
...
+		addr.Port, eof = source.NextUint16()
+		addr.ID, eof = source.NextUint64()
+		if eof {
+			return io.ErrUnexpectedEOF
+		}
 
+		this.NodeAddrs = append(this.NodeAddrs, addr)
+	}
+
+	if count > comm.MAX_ADDR_NODE_CNT {
```
BooniesFX: count exceed attack check should  before node address appending

p2pserver/message/types/address.go
```diff
@@ -32,44 +33,70 @@ type Addr struct {
...
+		count = comm.MAX_ADDR_NODE_CNT
+	}
+	this.NodeAddrs = this.NodeAddrs[:count]
+
+	return nil
+}
+
+//Deserialize message payload
+func (this *Addr) Deserialize(r io.Reader) error {
```
BooniesFX: remove Deserialize ??

p2pserver/message/types/inventory.go
```diff
@@ -49,47 +45,40 @@ func (this *Inv) CmdType() string {
...
+		hash, eof := source.NextHash()
+		if eof {
+			return io.ErrUnexpectedEOF
 		}
-		this.P.Blk = append(this.P.Blk, blk)
+
+		this.P.Blk = append(this.P.Blk, hash)
+	}
+
+	if blkCnt > p2pCommon.MAX_INV_BLK_CNT {
```
BooniesFX, line 79: check before blk append??

BooniesFX (changes requested):

p2pserver/message/types/address.go
laizy: 1.  the param is not `[]byte` any more, so all bytes belong to this message should be read out from the `source`, to avoid subtle bug. 2. the max payload size has checked, so this is safe.

BooniesFX: so this is safe now, should we remove this count limit??

p2pserver/message/types/inventory.go
laizy: same reason as above

p2pserver/message/types/address.go
laizy: since there is a global max outbound connection limit, the MAX_ADDR_NODE_CNT is to avoid  malicious node send too many addresses and occupy the outbound connection?

lightshine001 (approved):

PR #606, opened 2018-08-13
implement zero copy source/sink

laizy:
the current serialization has very poor performance caused by too many memory allocation.

### DIFF
--- a/common/zero_copy_sink.go
+++ b/common/zero_copy_sink.go
@@ -1,0 +1,152 @@
+package common
+
+import (
+	"bytes"
+	"encoding/binary"
+	"errors"
+)
+
+type ZeroCopySink struct {
+	buf []byte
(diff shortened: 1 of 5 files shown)


laizy:
benchmark:
serialize: 2.6x faster
```
BenchmarkSerialize-4                        5000            262269 ns/op          128032 B/op       8000 allocs/op
BenchmarkZeroCopySink-4                    10000            100427 ns/op              15 B/op          0 allocs/op
```
deserialize: 14x faster
```
BenchmarkZeroCopySource-4                1000000              1463 ns/op               0 B/op          0 allocs/op
BenchmarkDeserialize-4                     100000             20667 ns/op            7792 B/op        601 allocs/op
```

gasby88 (approved):

PR #609, opened 2018-08-16
modify for support array state param for transfer

JasonZhouPW:
since python compiler doesn't support Struct, we need to support array for transfer

### DIFF
--- a/smartcontract/service/native/ont/states.go
+++ b/smartcontract/service/native/ont/states.go
@@ -85,7 +85,7 @@ func (this *State) Deserialize(r io.Reader) error {
 		//for array case, the first element is count
 		//so from is the second element
 		this.From, err = utils.ReadAddress(r)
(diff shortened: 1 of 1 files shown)


smartcontract/service/native/ont/states.go
```diff
@@ -82,7 +82,12 @@ func (this *State) Deserialize(r io.Reader) error {
 	var err error
 	this.From, err = utils.ReadAddress(r)
 	if err != nil {
-		return fmt.Errorf("[State] deserialize from error:%v", err)
+		//for array case, the first element is count
+		//so from is the second element
+		this.From, err = utils.ReadAddress(r)
```
laizy, line 87: Can you elaborate when this will happen? ambiguous serialization format is not acceptable.

laizy:
close this because ambiguous serialization format is not acceptable.

PR #622, opened 2018-08-23
add struct clone length limit

tanZiWen:


vm/neovm/types/struct.go
```diff
@@ -79,17 +80,18 @@ func (s *Struct) GetStruct() ([]StackItems, error) {
...
+func clone(s *Struct, length int) (StackItems, error) {
+	if length > MAX_CLONE_LENGTH {
+		return nil, fmt.Errorf("%s", "over max struct clone length")
+	}
 	var arr []StackItems
 	for _, v := range s._array {
+		length++
 		if value, ok := v.(*Struct); ok {
-			vc, err := clone(value)
+			vc, err := clone(value, length)
```
laizy: buggy implementation. `length` should be `*int`

laizy (approved):

PR #638, opened 2018-09-03
 fix convert vm type to hexstring bug

tanZiWen:


smartcontract/service/neovm/config.go
```diff
@@ -46,7 +46,7 @@ var (
 
 	PER_UNIT_CODE_LEN   int = 1024
 	METHOD_LENGTH_LIMIT int = 1024
-	MAX_STACK_SIZE      int = 1024
+	MAX_STACK_SIZE      int = 1024 * 2
```
laizy: why need change this?

tanZiWen: EvaluationStack + AltStack = 2*1024

laizy (approved):

PR #641, opened 2018-09-04
[ONTO-481] Fix account duplicate import at sigsvr import cmd

gasby88:


cmd/sigsvr/store/wallet_store.go
```diff
@@ -257,7 +257,16 @@ func (this *WalletStore) NewAccountData(typeCode keypair.KeyType, curveCode byte
 	return accData, nil
 }
 
-func (this *WalletStore) AddAccountData(accData *account.AccountData) error {
+func (this *WalletStore) AddAccountData(accData *account.AccountData) (bool, error) {
+	data, err := json.Marshal(accData)
```
laizy: `accData` will be modified after marshal. it is intentional?

laizy (approved):

PR #655, opened 2018-09-18
ONT-1321 add QoS services of blocksync

chenzhijie:
* Replace NodeWeight with QoS service
* Add peer's throughput, capacity calculation
* Calculate TTL of flight more accurately
* Add height-first sync mode
* Allow request multiple blocks from same node by tasks management


Honglei-Cong:
question: any document for the feature?

PR #658, opened 2018-09-19
add addresstobase58,base58toaddress API support

tanZiWen:


common/address.go
```diff
@@ -118,3 +105,24 @@ func AddressFromBase58(encoded string) (Address, error) {
 
 	return ph, nil
 }
+
+func VerifyBase58(encoded string) ([]byte, error) {
```
laizy: the original `AddressFromBase58` has verified this, so any reason you need this function? and this implementation is buggy since the checksum is not verified.

tanZiWen: sorry, what's the meaning of 'buggy'

common/address.go
```diff
@@ -98,13 +98,18 @@ func AddressFromBase58(encoded string) (Address, error) {
...
-		return ADDRESS_EMPTY, fmt.Errorf("invalid address")
+		return ADDRESS_EMPTY, errors.New("invalid address")
 	}
 
 	buf := x.Bytes()
 	if len(buf) != 1+ADDR_LEN+4 || buf[0] != byte(23) {
 		return ADDRESS_EMPTY, errors.New("wrong encoded address")
 	}
+
+	if err != nil {
```
laizy: `err` has checked at line 95

core/store/ledgerstore/ledger_store.go
```diff
@@ -845,12 +844,18 @@ func (this *LedgerStoreImp) GetEventNotifyByBlock(height uint32) ([]*event.Execu
...
+	header, err := this.GetHeaderByHeight(height)
+	if err != nil {
+		return stf, err
+	}
 
 	config := &smartcontract.Config{
-		Time:   uint32(time.Now().Unix()),
-		Height: this.GetCurrentBlockHeight() + 1,
-		Tx:     tx,
+		Time:       header.Timestamp,
```
laizy: why revert this back?

laizy (dismissed):

PR #664, opened 2018-09-26
fix the panic issue in link.Rx()

fuwei1217:


p2pserver/link/link.go
```diff
@@ -110,30 +110,32 @@ func (this *Link) GetRXTime() time.Time {
 }
 
 func (this *Link) Rx() {
-	reader := bufio.NewReaderSize(this.conn, common.MAX_BUF_LEN)
+	if this.conn != nil {
+		reader := bufio.NewReaderSize(this.conn, common.MAX_BUF_LEN)
```
laizy: save `this.conn` to a local variable first to avoid data race problem? and return directly if conn==nil to avoid another calling of `disconnectNotify`

fuwei1217: code updated, please check

laizy (approved):

PR #686, opened 2018-11-09
共 --> 共识

lcgogo:


qiluge (approved):
Thank you. But the latest command line arguments is added '-' to split different word, such as ```--enable-consensus```, not ```--enableconsensus```. And there some alike mistakes in our documents, we will fix these mistakes very soon.

laizy:
fixed in #695

PR #705, opened 2018-12-08
Update ontology-crypto

WyattMufson:
The current ontology-crypto repo was updated, vendor currently points to an old commit. 

I updated the submodule to master.

Signed-off-by: Wyatt Mufson wyatt@ryucoin.com


AlverLyu:
It is managed with glide now. The sub-module should be removed. Sorry for the confusion.

laizy:
fixed in #706

PR #711, opened 2018-12-17
remove unused ontid interfaces

AlverLyu:


Honglei-Cong:
I take this as one OntID system contract upgrading.
System contract-upgrading process should be followed for this change.

laizy:
shadow chain is not needed any more.

PR #720, opened 2018-12-25
fix syncing stopped when network failed

Honglei-Cong:
syncing failed to catch up when network recovered from partitioned.
when network failed, syncer failed to retry.
In the fix, state-mgr compares maxCommitted with syncer target blocknum.  If syncer failed, state-mgr will restart the syncer.


Honglei-Cong:
still in test

PR #730, opened 2019-01-22
fix block root calculation

laizy:


consensus/vbft/msg_builder.go
```diff
@@ -183,7 +183,7 @@ func (self *Server) constructBlock(blkNum uint32, prevBlkHash common.Uint256, tx
 		txHash = append(txHash, t.Hash())
 	}
 	txRoot := common.ComputeMerkleRoot(txHash)
-	blockRoot := ledger.DefLedger.GetBlockRootWithNewTxRoot(txRoot)
+	blockRoot := ledger.DefLedger.GetBlockRootWithNewTxRoots(txRoot)
```
Honglei-Cong: txRoot of last-block should also be included here?

core/ledger/ledger.go
```diff
@@ -84,8 +84,8 @@ func (self *Ledger) GetStateMerkleRoot(height uint32) (result common.Uint256, er
 	return self.ldgStore.GetStateMerkleRoot(height)
 }
 
-func (self *Ledger) GetBlockRootWithNewTxRoot(txRoot common.Uint256) common.Uint256 {
-	return self.ldgStore.GetBlockRootWithNewTxRoot(txRoot)
+func (self *Ledger) GetBlockRootWithNewTxRoots(txRoots []common.Uint256) common.Uint256 {
```
Honglei-Cong: suggest changing this interface to
GetBlockRootWithNewTxRoots(txRootHeight uint32, newTxRoots []common.Uint256)
： newTxRoots includes merkle roots starting from block txRootHeight

when calculating the new root, ledger should check if any txRoots missed or inconsistent.

consensus/vbft/msg_builder.go
laizy: yes, this is left for @xiemylogos to implement. so the title contains `[wip]`.

core/ledger/ledger.go
laizy: en, this is what I am processing to make the api stateless, which is necessary since the ledger is accessed concurrently

consensus/vbft/msg_builder.go
xiemylogos: ok, I am modifying

Honglei-Cong (approved):

PR #755, opened 2019-02-19
include version 1.5.3 code 

laizy:
* fix main net fee split bug
* add address check when authorize


jonatanblockchain:
spelling
//check if address is **peer** owner address

tanZiWen (approved):

PR #758, opened 2019-02-22
Fix the sync node crash problem when vast nodes connect to it

likaioc:


Honglei-Cong:
please gofmt.

p2pserver/message/utils/msg_router.go
```diff
@@ -57,6 +59,11 @@ func (this *MessageRouter) init(p2p p2p.P2P) {
 	this.stopConsCh = make(chan bool)
 	this.p2p = p2p
 
+	this.msgWP = &msgWorkerPool{
+		maxWorkerCount : 256 * config.DefConfig.P2PNode.MaxConnInBoundForSingleIP,
```
Honglei-Cong: too much?

p2pserver/message/utils/msg_workerpool.go
```diff
@@ -0,0 +1,232 @@
...
+	"sync"
+	"time"
+
+	"github.com/ontio/ontology-eventbus/actor"
+	"github.com/ontio/ontology/common/log"
+	"github.com/ontio/ontology/p2pserver/message/types"
+	"github.com/ontio/ontology/p2pserver/net/protocol"
+)
+
+type msgJobChanList *[]*msgJobChan
```
Honglei-Cong, line 51: ???

p2pserver/message/utils/msg_workerpool.go
```diff
@@ -0,0 +1,232 @@
...
+	msgHandler MessageHandler
+}
+
+type msgJobChan struct {
+	lastUseTime time.Time
+	jobChan     chan *msgJobItem
+}
+
+func (this* msgWorkerPool) init() {
+
```
Honglei-Cong, line 83: remove these un-necessary blank lines?

p2pserver/message/utils/msg_workerpool.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	return msgJobCh
+}
+
+func (this* msgWorkerPool) receiveMsg(msg *msgJobItem) bool {
+
+	msgJobCh:= this.getMsgWorkChan(msg.msgPayload.Payload.CmdType())
+	if msgJobCh != nil {
+		msgJobCh.jobChan <- msg
```
Honglei-Cong, line 214: This function should be lock-free.
drop msg if jobChan is full.

Honglei-Cong:
should WorkerPool drop msg when there's no idle worker?

p2pserver/message/utils/msg_workerpool.go
```diff
@@ -0,0 +1,225 @@
...
+func (this *msgWorkerPool) start() {
+
+	if this.waitingWokers == nil || this.stopChan == nil {
+		log.Error("[p2p]invalid start invoking, the msg worker pool hasn't been initialized")
+		return
+	}
+
+	go func() {
+		for {
+			this.clean()
```
Honglei-Cong, line 111: to be simple, why not start all workers here, and let them waiting on one channel?

Honglei-Cong (changes requested):

laizy:
do not force push after some one has reviewed.

p2pserver/message/utils/msg_router.go
likaioc: 256 is the goroutine numbers for single inbound peer. In my opion,  goroutine  is minor executing unit, 256 is not too much for single peer.

p2pserver/message/utils/msg_workerpool.go
likaioc: The is a slice pointer, the slice element is a pointer.  A  better way for this to provide some comon libs for these basic data struct in the whole platform .

p2pserver/message/utils/msg_workerpool.go
likaioc: Yes, have removed

p2pserver/message/utils/msg_workerpool.go
```diff
@@ -0,0 +1,224 @@
...
+
+	"github.com/ontio/ontology-eventbus/actor"
+	"github.com/ontio/ontology/common/log"
+	"github.com/ontio/ontology/p2pserver/message/types"
+	"github.com/ontio/ontology/p2pserver/net/protocol"
+)
+
+type msgJobChanList *[]*msgJobChan
+
+type msgWorkerPool struct {
```
laizy, line 53: this worker pool implementation looks familiar to me. If I remember correctly, it is copied and modified from `fasthttp`, so add original copyright in this file.

laizy (changes requested):

p2pserver/message/utils/msg_workerpool.go
likaioc: You are right. I have modify it.

p2pserver/message/utils/msg_workerpool.go
likaioc: If the peer connection count is smaller, if all workers are waiting,  the resources are waste.

p2pserver/message/utils/msg_workerpool.go
likaioc: Yes, the worker pool is modified based on fasthttp. I have added copyright of fasthttp.

p2pserver/message/utils/msg_workerpool.go
laizy: adjust to `func NewWorkerPool() *msgWrokerPool`, constructing struct using two step is really ugly.

p2pserver/message/utils/msg_workerpool.go
```diff
@@ -0,0 +1,246 @@
...
+				return
+			default:
+				time.Sleep(this.maxWorkerIdleTime)
+			}
+		}
+	}()
+}
+
+func (this *msgWorkerPool) stop() {
+
```
laizy: why empty line in the start of each function body?

p2pserver/message/utils/msg_workerpool.go
```diff
@@ -0,0 +1,246 @@
...
+	// since otherwise the receiveMsg caller (Acceptor) may lag accepting
+	// new message if disposeMsg is CPU-bound.
+	return 1
+}()
+
+func (this *msgWorkerPool) start() {
+
+	if this.waitingWokers == nil || this.stopChan == nil {
+		log.Error("[p2p]invalid start invoking, the msg worker pool hasn't been initialized")
+		return
```
laizy: why return silently? the original implementation is sound, I don't understand why you need to modify it.

p2pserver/message/utils/msg_workerpool.go
laizy: I think slice is enough, why need a slice pointer?

p2pserver/message/utils/msg_workerpool.go
laizy: I don't think you get the point. `receiveMsg` can not be blocked since it will cause `MessageRouter` stall. what if `jobChan` is full?

p2pserver/message/utils/msg_workerpool.go
```diff
@@ -0,0 +1,253 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with The ontology.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*
+ * https://github.com/valyala/fasthttp
+ * The MIT License (MIT)
+
+ *Copyright (c) 2015-present Aliaksandr Valialkin, VertaMedia
```
Honglei-Cong, line 23: If we have to copy other's code, I suggest add them as 3rd party vendor package, rather than copy/paste.
What you think?

likaioc: The origin's implement is not common and can't directly apply to this context.

p2pserver/message/utils/msg_workerpool.go
likaioc: From the point of function,  slice is enough. The purpose of slice pointer is to reduce the copy and to be convenient for operation.

p2pserver/message/utils/msg_workerpool.go
likaioc: The msg woker pool's purpose is to limit vast msg specific disposing.  If the block sync msg is not disposed , the MessageRouter only discard it. In my opion, this won't cause MessageRouter stall.
     In receiveMsg, once all jobChan are using,  getMsgWorkChan will return nil and it won't affect MessageRouter  disposing other msgs.

p2pserver/message/utils/msg_workerpool.go
likaioc: My purpose is to let function and function body clear at a glance.

p2pserver/message/utils/msg_workerpool.go
likaioc: Yes,  I have adjusted.

p2pserver/message/utils/msg_workerpool.go
likaioc: My original purpose is to avoid the panic once error initialization happen and can't affect other disposings. But the panic has other good aspect. However, I have changed to panic.

p2pserver/message/utils/msg_workerpool.go
laizy: I think you need following the coding conventions.

likaioc: Ok， I remove the empty.

PR #760, opened 2019-02-26
set ulimit to increase the capacity of open files

laizy:


core/store/leveldbstore/leveldb_store.go
```diff
@@ -19,6 +19,7 @@
 package leveldbstore
 
 import (
+	"github.com/ethereum/go-ethereum/common/fdlimit"
```
Honglei-Cong, line 22: update glide config?

core/store/leveldbstore/leveldb_store.go
```diff
@@ -40,11 +41,21 @@ const BITSPERKEY = 10
 
 //NewLevelDBStore return LevelDBStore instance
 func NewLevelDBStore(file string) (*LevelDBStore, error) {
+	openFileCache := 500
```
Honglei-Cong: define as const?

Honglei-Cong (approved):

PR #774, opened 2019-03-06
improve broadcast performance

Honglei-Cong:
1. serialize before transmitting on peer link
2. improve msg header deserialization
3. change Send(msg) to Send(msgType, msgPayload), add SendMsg(msg) interface

in broadcasting, firstly serialize msg to []byte, then transmitting on each peer link with Send(payload)

### DIFF
--- a/p2pserver/link/link.go
+++ b/p2pserver/link/link.go
@@ -25,7 +25,6 @@ import (
 	"net"
 	"time"
 
-	comm "github.com/ontio/ontology/common"
 	"github.com/ontio/ontology/common/log"
 	"github.com/ontio/ontology/p2pserver/common"
 	"github.com/ontio/ontology/p2pserver/message/types"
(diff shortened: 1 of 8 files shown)


p2pserver/message/types/message.go
```diff
@@ -53,8 +52,20 @@ type messageHeader struct {
 
 func readMessageHeader(reader io.Reader) (messageHeader, error) {
 	msgh := messageHeader{}
-	err := binary.Read(reader, binary.LittleEndian, &msgh)
-	return msgh, err
+	hdrBytes := make([]byte, comm.UINT32_SIZE+common.MSG_CMD_LEN+comm.UINT32_SIZE+common.CHECKSUM_LEN)
+
+	if _, err := reader.Read(hdrBytes); err != nil {
```
laizy: it is a common mistake to use `Reader.Read`, need use `io.ReadFull`

p2pserver/link/link.go
```diff
@@ -166,20 +165,12 @@ func (this *Link) CloseConn() {
 	}
 }
 
-func (this *Link) Tx(msg types.Message) error {
+func (this *Link) Tx(msgType string, payload []byte) error {
```
laizy: unused `msgType` .  the original method is convenient to call. rename it to `Send` and add another method ?
```golang
Send(msg types.Message) error // for common usage
SendRaw(rawPacket  []byte) error // for broadcast, avoid serialization many times 

```

p2pserver/message/types/message.go
Honglei-Cong: ok

laizy (approved):

Issue #806, opened 2019-03-12
Consensus halt


Krestol:
1.6.0-beta

The issue is actual for the latest release
The issue reproduces on Windows and Linux

Steps:
1. run 7 consensus nodes
2. suspend 4 of 7 consensus nodes
3. resume all suspended nodes

Expected Result :
New blocks appear in the blockchain

Actual Result:
Consensus halt is present. Now new blocks appear  in the blockchain

For running private network on Windows you can unpack archive [private_network.zip](https://github.com/ontio/ontology/files/2957943/private_network.zip), copy main.exe in it and run launch.cmd

xiemylogos:
Hi Krestol.can you send your test log,I test 1.6.0-beta use your method, It looks like everything is ok. I review your https://github.com/ontio/ontology/pull/807 .but  There are some errors, use this pr in new consensus code ,the consensus halt. because your function getConsensusedCommittedBlockNum()   always return 0,false.

Honglei-Cong:
fixed by #807

PR #807, opened 2019-03-12
Fix Consensus halt #806

Krestol:
The misbehavior of this function is the reason why consensus may be halted

Signed-off-by: Krestol <krestol@i.ua>


consensus/vbft/state_mgmt.go
```diff
@@ -428,27 +429,34 @@ func (self *StateMgr) checkStartSyncing(startBlkNum uint32, forceSync bool) erro
...
-			}
-			for k := range peers {
-				if n >= k {
-					peers[k] = append(peers[k], p.peerIdx)
-				}
-			}
-			if len(peers[n]) > C {
-				maxCommitted = n
-				consensused = true
+	for k := range commitedBlocks {
```
Honglei-Cong: for simple, change to the following?

if len(committedBlocks) <= C || committedBlock[C] < myCommitted { 
    return myCommitted, false 
} else {
    return  committedBlock[C], true
}

consensus/vbft/state_mgmt.go
```diff
@@ -428,27 +429,34 @@ func (self *StateMgr) checkStartSyncing(startBlkNum uint32, forceSync bool) erro
...
+
+	commitedBlocksList := make([]uint32, 0, len(self.peers))
+	for _, p := range self.peers {
+		if p.committedBlockNum > myCommitted {
+			commitedBlocksList = append(commitedBlocksList, p.committedBlockNum)
+		}
+	}
 
+	sort.Slice(commitedBlocksList, func(i, j int) bool { return commitedBlocksList[i] > commitedBlocksList[j] })
+
```
Honglei-Cong: for simple, change to the following?

```
if len(committedBlocksList) <= C || committedBlocksList[C] < myCommitted { 
    return myCommitted, false 
} else {
    return  committedBlocksList[C], true
}
```

Krestol: @Honglei-Cong 
1. commitedBlockslist is a list of committed blocks.
len(committedBlocksList) - returns number of peers, so we can't compare it with the quorum value

2. committedBlocksList[C] - number of committed block of one of the peers, e.x. if quorum value is 2, than committedBlocksList[C] returns committed block of the third peer.

Honglei-Cong: Right.  After commitedBlockList was sorted by order of height, committedBlocksList[C] is the consensused block height?

Krestol: Yes, after commitedBlockList is sorted it has sens. I'll do these changes.

Honglei-Cong (approved):
looks good to me.

PR #808, opened 2019-03-15
Add a new native contract to support ec curve operations and zk-SNARK verification

kunxian-xia:
This new contract has six functions: 
- ECAdd(): add two ec points on the `alt_bn128` curve
- TwistECAdd(): add two ec points on the twist curve
- ECMul(): scalar multiplication on the `alt_bn128` curve
- TwistECMul(): scalar multiplication on the twist curve
- PairingCheck(): check if `e(a1, b1) ... e(ak, bk) == 1`  where `ai` are points on the `alt_bn128` curve and `bi` are points on the twist curve. 
- PHGR13Verify(vk, proof, input): verify if a proof is a valid PHGR13 zk-SNARKs proof corresponding to the vk.


Honglei-Cong:
fixes https://github.com/ontio/ontology/issues/810

Honglei-Cong:
@kunxian-xia 
Is there any project requiring this feature?
If not, suggest to commit to branch shard-common.
What you think? @kunxian-xia @laizy

PR #811, opened 2019-03-18
fix contract migrate key error

JasonZhouPW:
1. fix contract migrate key error
2. notify function return string object

### DIFF
--- a/smartcontract/service/wasmvm/contract.go
+++ b/smartcontract/service/wasmvm/contract.go
@@ -182,7 +182,7 @@ func (self *Runtime) ContractMigrate(proc *exec.Process,
 		key := iter.Key()
 		val := iter.Value()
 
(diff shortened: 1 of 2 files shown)


smartcontract/service/wasmvm/contract.go
```diff
@@ -182,7 +182,7 @@ func (self *Runtime) ContractMigrate(proc *exec.Process,
 		key := iter.Key()
 		val := iter.Value()
 
-		newkey := serializeStorageKey(contractAddr, key)
+		newkey := serializeStorageKey(contractAddr, key[20:])
```
Honglei-Cong, line 185: what is 20?

laizy: length of contract Address

laizy (approved):

PR #816, opened 2019-03-21
Create pledge.py

chariotsfanatics:
DAD token pledge


Honglei-Cong:
Should this request for https://github.com/DAD-Chain/ontology?

PR #821, opened 2019-03-25
optimize consensus code,modify return type when no error msg

xiemylogos:


consensus/vbft/event_timer.go
```diff
@@ -190,7 +187,7 @@ func (self *EventTimer) getEventTimeout(evtType TimerEventType) time.Duration {
 //
 // internal helper, should call with lock held
 //
-func (self *EventTimer) startEventTimer(evtType TimerEventType, blockNum uint32) error {
+func (self *EventTimer) startEventTimer(evtType TimerEventType, blockNum uint32) {
```
Honglei-Cong: suggest changing the panic to 'return err'

Honglei-Cong (commented):
overall, looks good to me.
except one case: err should be returned on 'start timer'

consensus/vbft/event_timer.go
xiemylogos: ok

Honglei-Cong (approved):
looks good.

PR #823, opened 2019-03-26
fix bigint to neo bytes error when int is 0

laizy:


Honglei-Cong (commented):
looks good.

laizy:
It is a shame that this fix will cause hard fork on mainnet, so close this.

PR #826, opened 2019-03-27
fix vbft shard user stake and seal block bug

qiluge:
1. fix stake param shard id serialize bug;
2. update vbft parent height at each proposal;
3. only update user unbound ong info while stake asset is ONT contract

### DIFF
--- a/smartcontract/service/native/shard_stake/params.go
+++ b/smartcontract/service/native/shard_stake/params.go
@@ -204,12 +204,12 @@ func (this *UnfreezeFromShardParam) Deserialize(r io.Reader) error {
 }
 
 type WithdrawStakeAssetParam struct {
-	ShardId types.ShardID
+	ShardId uint64
 	User    common.Address
 }
 
 func (this *WithdrawStakeAssetParam) Serialize(w io.Writer) error {
-	if err := utils.WriteVarUint(w, this.ShardId.ToUint64()); err != nil {
+	if err := utils.WriteVarUint(w, this.ShardId); err != nil {
 		return fmt.Errorf("serialize: write shard id failed, err: %s", err)
 	}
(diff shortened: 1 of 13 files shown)


smartcontract/service/native/shard_stake/params.go
```diff
@@ -189,12 +204,12 @@ func (this *UnfreezeFromShardParam) Deserialize(r io.Reader) error {
 }
 
 type WithdrawStakeAssetParam struct {
-	ShardId types.ShardID
+	ShardId uint64
```
laizy: why modify this?

smartcontract/service/native/shard_stake/params.go
```diff
@@ -208,7 +223,7 @@ func (this *WithdrawStakeAssetParam) Deserialize(r io.Reader) error {
 	if err != nil {
 		return fmt.Errorf("deserialize: read shard id failed, err: %s", err)
 	}
-	this.ShardId, err = types.NewShardID(id)
+	this.ShardId = id
```
laizy, line 226: not all uint64 are valid ShardID, the original check is needed.

laizy (changes requested):

smartcontract/service/native/shard_stake/params.go
```diff
@@ -421,14 +463,14 @@ func (this *UserStakeParam) Deserialize(r io.Reader) error {
 }
 
 type ChangeMaxAuthorizationParam struct {
-	ShardId    types.ShardID
+	ShardId    uint64
```
laizy: revert all those changes back

qiluge:
@laizy fixed shard serialize problem

Honglei-Cong (approved):

PR #834, opened 2019-04-02
optimize neovm

laizy:
- [x] implement optimization
- [x] opcode execution compare test with original implementation
- [x] ledger consistence check with mainnet
- [x] coverage test


core/store/ledgerstore/ledger_store.go
```diff
@@ -630,11 +629,19 @@ func (this *LedgerStoreImp) executeBlock(block *types.Block) (result store.Execu
 			return
 		}
 	}
+	gasTable := make(map[string]uint64)
+	neovm.GAS_TABLE.Range(func(k, value interface{}) bool {
```
Honglei-Cong, line 637: why not change neovm.GAS_TABLE from sync.map to map with RWLock?

laizy: from the implementation perspective, there is little difference between `sync.map` and `map with RWLock`. clone a local map is a cheap way to avoid locking before each access.

laizy:
randomly generate 2000w opcode snippets, the execution results of two vms are equal. so I think it is ready to review now.

core/store/ledgerstore/ledger_store.go
carltraveler: why need sync.map. is there any prossible that  GAS_TABLE have data race?

smartcontract/common/common.go
```diff
@@ -33,12 +32,12 @@ const (
 	MAX_COUNT = 1024
 )
 
-func ConvertNeoVmTypeHexString(item interface{}) (interface{}, error) {
+func ConvertNeoVmTypeHexString(item types.StackItems) (interface{}, error) {
```
carltraveler, line 35: this func return string or array. maybe it can patch like *type.vmvalue to improve performance?

smartcontract/service/native/native_service.go
```diff
@@ -59,16 +58,16 @@ func (this *NativeService) Register(methodName string, handler Handler) {
 	this.ServiceMap[methodName] = handler
 }
 
-func (this *NativeService) Invoke() (interface{}, error) {
+func (this *NativeService) Invoke() ([]byte, error) {
 	contract := this.InvokeParam
 	services, ok := Contracts[contract.Address]
 	if !ok {
-		return false, fmt.Errorf("Native contract address %x haven't been registered.", contract.Address)
+		return BYTE_FALSE, fmt.Errorf("Native contract address %x haven't been registered.", contract.Address)
```
carltraveler, line 65: this restricted the native service must return []byte. maybe currently only return bool, string, big.int. however, if someday need add a native service return array. this code will convert back. can it replace with vmvalue?

smartcontract/service/neovm/contract.go
```diff
@@ -132,64 +132,65 @@ func ContractGetStorageContext(service *NeoVmService, engine *vm.ExecutionEngine
...
-	i, err := vm.PopInteropInterface(engine)
+func ContractGetCode(service *NeoVmService, engine *vm.Executor) error {
+	i, err := engine.EvalStack.PopAsInteropValue()
 	if err != nil {
 		return err
 	}
-	vm.PushData(engine, i.(*payload.DeployCode).Code)
-	return nil
+	if d, ok := i.Data.(*payload.DeployCode); ok {
+		return engine.EvalStack.PushBytes(d.Code)
```
carltraveler, line 145: PushBytes restricted the bytearray size MAX_BYTEARRAY_SIZE which is defined to 1024*1024. I do not search the maro MAX_BYTEARRAY_SIZE when deploy code.

smartcontract/service/neovm/contract.go
```diff
@@ -132,64 +132,65 @@ func ContractGetStorageContext(service *NeoVmService, engine *vm.ExecutionEngine
...
+func isContractParamValid(engine *vm.Executor) (*payload.DeployCode, error) {
+	if engine.EvalStack.Count() < 7 {
 		return nil, errors.NewErr("[Contract] Too few input parameters")
 	}
-	code, err := vm.PopByteArray(engine)
+	code, err := engine.EvalStack.PopAsBytes()
 	if err != nil {
 		return nil, err
 	}
 	if len(code) > 1024*1024 {
```
carltraveler, line 158: here len(code) > 1024*1024 need use MAX_BYTEARRAY_SIZE  better.

smartcontract/service/neovm/neovm_service.go
```diff
@@ -123,101 +117,76 @@ type NeoVmService struct {
...
 		case vm.APPCALL:
 			address, err := this.Engine.Context.OpReader.ReadBytes(20)
 			if err != nil {
-				return nil, fmt.Errorf("[Appcall] read contract address error: %v", err)
+				return nil, fmt.Errorf("[Appcall] read contract address error:%v", err)
 			}
-			if bytes.Compare(address, BYTE_ZERO_20) == 0 {
-				if vm.EvaluationStackCount(this.Engine) < 1 {
-					return nil, fmt.Errorf("[Appcall] too few input parameters: %d", vm.EvaluationStackCount(this.Engine))
+			if bytes.Compare(address, scommon.ADDRESS_EMPTY[:]) == 0 {
```
carltraveler, line 180: so the uninitial global var ADDRESS_EMPTY must be cleared by 0 when go init?

smartcontract/service/neovm/runtime_test.go
```diff
@@ -19,41 +19,68 @@
...
-	result, err := item.GetBigInteger()
+	in_t := types.NewInteger(b)
+	m_old := types.NewMap()
+	m_old.Add(types.NewByteArray([]byte("key")), in_t)
+	bss, err := SerializeStackItem(m_old)
 	assert.Nil(t, err)
-
-	assert.Equal(t, result, i)
-
+	assert.Equal(t, "820100036b65790209ffffc58e4ae6b68900", common.ToHexString(bss))
```
carltraveler, line 86: may be here can add some test about the coherence of the old serialize/deserilize method and the vmvalue.serialize/deserialize method

smartcontract/service/neovm/runtime_test.go
```diff
@@ -126,52 +153,3 @@ func TestStructRef(t *testing.T) {
...
-
-	err = RuntimeAddressToBase58(nil, vm)
-
-	assert.NoError(t, err)
-
-	result, err := vm.EvaluationStack.Pop().GetByteArray()
-
-	assert.NoError(t, err)
-	assert.Equal(t, base58, string(result))
-}
```
carltraveler: why base58 test deleted?

smartcontract/test/common_test.go
```diff
@@ -50,3 +51,35 @@ func TestConvertNeoVmTypeHexString(t *testing.T) {
...
+			Gas:    1000000,
+		}
+		engine, err := sc.NewExecuteEngine(code)
+		if err != nil {
+			panic(err)
+		}
+		_, err = engine.Invoke()
+		if err != nil {
+			panic(err)
+		}
```
carltraveler, line 82: curious how's the performance?

vm/neovm/executor.go
```diff
@@ -0,0 +1,1101 @@
...
+			if err != nil {
+				return FAULT, err
+			}
+			numBytes = int(num)
+		} else {
+			num, err := context.OpReader.ReadUint32()
+			if err != nil {
+				return FAULT, err
+			}
+			numBytes = int(num)
```
carltraveler: num is uint32, but here converted to int

carltraveler: and what strange is that the max bytearray size is 1024*1024, so what's the point of PUSHDATA4. PUSHDATA2 is enough

smartcontract/service/neovm/neovm_service.go
```diff
@@ -123,101 +117,76 @@ type NeoVmService struct {
...
 	PreExec       bool
 }
 
 // Invoke a smart contract
-func (this *NeoVmService) Invoke() (interface{}, error) {
+func (this *NeoVmService) Invoke() (*vmty.VmValue, error) {
 	if len(this.Code) == 0 {
 		return nil, ERR_EXECUTE_CODE
 	}
 	this.ContextRef.PushContext(&context.Context{ContractAddress: scommon.AddressFromVmCode(this.Code), Code: this.Code})
```
carltraveler, line 134: did the ContextRef PushContext check the deth, or if A appcall B. and B appcall A. this will cause dead loop

vm/neovm/params.go
```diff
@@ -19,9 +19,9 @@
 package neovm
 
 const (
-	STACK_LIMIT               uint32 = 2 * 1024
+	STACK_LIMIT                      = 2 * 1024
 	MAX_INVOCATION_STACK_SIZE        = 1024
 	MAX_SIZE_FOR_BIGINTEGER          = 32
 	MAX_BYTEARRAY_SIZE        uint32 = 1024 * 1024
```
carltraveler, line 25: why only MAX_BYTEARRAY_SIZE   have type def?

vm/neovm/executor.go
```diff
@@ -0,0 +1,1101 @@
...
+			return FAULT, errors.ERR_OVER_MAX_ARRAY_SIZE
+		}
+
+		b := arr[start:end]
+		err = self.EvalStack.PushBytes(b)
+		if err != nil {
+			return FAULT, err
+		}
+
+	case LEFT:
```
carltraveler, line 420: all bytearray sub slice can case []byte{} which is a blank bytearray. and how ever big.int between []byte{} and []byte{0} has no difference.  how ever in map keys []byte{} converted to string and []byte{0] converted to string are different keys. the sematic seems vague.

vm/neovm/executor.go
```diff
@@ -0,0 +1,1101 @@
...
+		}
+	case EQUAL:
+		left, right, err := self.EvalStack.PopPair()
+		if err != nil {
+			return FAULT, err
+		}
+		err = self.EvalStack.PushBool(left.Equals(right))
+		if err != nil {
+			return FAULT, err
+		}
```
carltraveler, line 513: also the EQUAL. []byte{} and []byte{0} not equal. howerver the internal big.int is equal

vm/neovm/types/map_value.go
```diff
@@ -0,0 +1,94 @@
...
+		return err
+	}
+
+	delete(this.Data, skey)
+
+	return nil
+}
+
+func (this *MapValue) Get(key VmValue) (value VmValue, ok bool, err error) {
+	skey, e := key.GetMapKey()
```
carltraveler, line 58: []byte{} []byte{0} []byte{0，0，0} ， so acctaully the boolean , big.int, bytearry still exploure to coder by some way. like equal compare by big.int or []byte have difference result. this  also can have difference sematic with old equal. the old equal's compare decided by the first argument type. like if the first one is interger and than the sencond one consider as interger too.

vm/neovm/types/neovm_value.go
```diff
@@ -0,0 +1,736 @@
...
+		return InteropValue{}, errors.ERR_BAD_TYPE
+	}
+}
+
+func (self *VmValue) Equals(other VmValue) bool {
+	v1, e1 := self.AsBytes()
+	v2, e2 := other.AsBytes()
+	if e1 == nil && e2 == nil { // both are primitive type
+		return bytes.Equal(v1, v2)
+	}
```
carltraveler, line 599: as mention above. equal difference with old one. like 0+0 is []byte{0,0}. the old one is equal. now not equal. becasueof the differenct lenght.

carltraveler (commented):
most important issue is the restriction of valstack. bytearry, call conext depth, and array index check.
a. the valstack length checker influence all  opcode pop num < push num. like unpack, any pushx op, TUCK, DUP, ALSTACK op. 
b. the bytearray length checker influence all opcode of slice, like cat, or big.int arithmatic which can be transfer to bytearray. so when do the big.int arithmatic also need restrict the same with array.
c. context deep checker influence the call and dcall.
d. array lenght checker influence append opocde. 

I think all these bound checker are necessary. however need be sure the hard fork.

smartcontract/common/common.go
laizy: This is legacy code, need to be removed finally.

> why need sync.map. is there any prossible that GAS_TABLE have data race?

transaction pre-execute will concurrently access the GAS_TABLE.

smartcontract/service/native/native_service.go
laizy: currently, the native contract is stable enough, it has little chance to upgrade in the near future.

smartcontract/service/neovm/contract.go
laizy: this check is redundant, since the code is got from vm, it should be checked there. In `code-audit-preview` branch, it has been removed.

smartcontract/service/neovm/neovm_service.go
laizy: golang will guarantee this. uninitialized value will have it's default value.

smartcontract/service/neovm/runtime_test.go
laizy: @lucas7788

vm/neovm/params.go
laizy: actually `golang` constant integer does not need type explicitly.

smartcontract/service/neovm/neovm_service.go
laizy: the original code checked this in `NewExecuteEngine`, so we keep it there. Anyway, I agree it is better to check it in `PushContext` to make code easier to reason.

vm/neovm/executor.go
laizy: @lucas7788

carltraveler (approved):

PR #835, opened 2019-04-02
use standard serialize at shard native contract

qiluge:
1. use neovm serialize at native contract param;
2. use zero copy at contract states and event;
3. fix build cross shard tx method;


Honglei-Cong:
Please resolve conflicts.

Honglei-Cong (approved):

PR #841, opened 2019-04-10
fix vbft consensus block sign number

xiemylogos:
1、when node receive endorser msg from node  the same as receive commiter msg from node
2、fix block proposer node not participate endorser and commiter @Honglei-Cong @laizy @tanZiWen

### DIFF
--- a/consensus/vbft/node_utils.go
+++ b/consensus/vbft/node_utils.go
@@ -334,8 +334,14 @@ func getCommitConsensus(commitMsgs []*blockCommitMsg, C int) (uint32, bool) {
 		}
 
 		commitCount[c.BlockProposer] += 1
-		if commitCount[c.BlockProposer] > C {
-			return c.BlockProposer, emptyCommit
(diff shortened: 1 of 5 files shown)


consensus/vbft/node_utils.go
```diff
@@ -334,8 +334,14 @@ func getCommitConsensus(commitMsgs []*blockCommitMsg, C int) (uint32, bool) {
 		}
 
 		commitCount[c.BlockProposer] += 1
-		if commitCount[c.BlockProposer] > C {
-			return c.BlockProposer, emptyCommit
+		if commitCount[c.BlockProposer] > C && commitCount[c.BlockProposer] == len(c.EndorsersSig) {
```
Honglei-Cong: This function can be improved by counting sigs of proposals in commit msgs.

Honglei-Cong (changes requested):

consensus/vbft/node_utils.go
xiemylogos: ok

Honglei-Cong (approved):
Looks good :)

PR #846, opened 2019-04-12
some zero copy serialization shouldn't return error

qiluge:


.gitignore
```diff
@@ -28,3 +28,6 @@ neosnapshot/*
 glide.lock
 *.iml
 docker/payload
+config.json
+peers.recent
+shard_*/
```
Honglei-Cong: suggest to remove shard_*/

laizy (approved):

Issue #737, opened 2019-01-28
block syncing consumed too much memory


Honglei-Cong:
Setting up syncing-node with 8GB memory, syncing node panic OOM after a few hours

Honglei-Cong:
In performance test, if too much block/header requests from peers, lots of memory are allocated.

Honglei-Cong:
https://github.com/ontio/ontology/pull/852

PR #852, opened 2019-04-18
add read back pressure in p2p link to avoid memory attack

laizy:


p2pserver/link/link.go
```diff
@@ -31,21 +30,37 @@ import (
 	"github.com/ontio/ontology/p2pserver/message/types"
 )
 
+const MaxWriteBuff = 16 * 1024 // add backpressure when pending msg in write buffer execced this
```
Honglei-Cong, line 33: 16K ？
PER_SEND_LEN is 256KB.

p2pserver/link/link.go
```diff
@@ -166,44 +198,75 @@ func (this *Link) CloseConn() {
...
-		log.Infof("[p2p]error sending messge to %s :%s", this.GetAddr(), err.Error())
-		this.disconnectNotify()
-		return err
+		conn.SetWriteDeadline(time.Now().Add(time.Duration(nCount*common.WRITE_DEADLINE) * time.Second))
+		_, err := conn.Write(payload)
+		if err != nil {
+			this.lock.Lock()
+			this.err = err
+			this.disconnectNotify()
+			this.cond.Signal() // notify read loop
```
Honglei-Cong, line 237: signal after unlock?

and make broadcast in disconnectNotify?

p2pserver/link/link.go
```diff
@@ -118,10 +138,25 @@ func (this *Link) Rx() {
...
 	for {
+		this.lock.Lock()
+		for this.err == nil && this.writeBuff.Size() > uint64(this.maxWriteBuff) {
+			this.cond.Wait()
+		}
+		if this.err != nil {
+			this.lock.Unlock()
+			break
+		}
+		this.lock.Unlock()
```
Honglei-Cong, line 149: move this line to Line145?

laizy: access `this.err` should be in locked context

p2pserver/link/link.go
laizy: there are just `readLoop` and `writeLoop` using this condition variable, since we are in `writeLoop`, only `readLoop` will wait for this, so `Signal` should be enough.

laizy: from go doc, it is ok to call `Singal` in locked context and there is little performance differance.

p2pserver/link/link.go
laizy: `16K` to start adding backpressure, but it still ok to call `Link.Send` to append more data to this buffer. And the final buffer size may execeed this value.

Honglei-Cong (approved):

Honglei-Cong:
https://github.com/ontio/ontology/issues/737

PR #858, opened 2019-04-23
adapt deploy payload to shard

qiluge:
1. modify deploy payload struct ;
2. support contract active and freeze.

### DIFF
--- a/smartcontract/service/neovm/config.go
+++ b/smartcontract/service/neovm/config.go
@@ -81,6 +81,8 @@ var (
 
 	CONTRACT_CREATE_NAME            = "Ontology.Contract.Create"
 	CONTRACT_MIGRATE_NAME           = "Ontology.Contract.Migrate"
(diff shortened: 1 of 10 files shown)


smartcontract/service/neovm/config.go
```diff
@@ -81,6 +81,8 @@ var (
 
 	CONTRACT_CREATE_NAME            = "Ontology.Contract.Create"
 	CONTRACT_MIGRATE_NAME           = "Ontology.Contract.Migrate"
+	CONTRACT_ACTIVE_NAME            = "Ontology.Contract.Active"
```
Honglei-Cong, line 84: Suggest to add new type of smart contract management transaction for activation/freeze.

These management transactions can only be processed on main-chain, because smart-contract metadata is stored in main-ledger.

Honglei-Cong (changes requested):

qiluge:
split meta data from deploy code

PR #863, opened 2019-04-25
reject block message with empty merkle root

laizy:


p2pserver/message/types/block.go
```diff
@@ -55,5 +57,10 @@ func (this *Block) Deserialization(source *common.ZeroCopySource) error {
 		// to accept old node's block
 		this.MerkleRoot = common.UINT256_EMPTY
 	}
+
+	stateHashHeight := config.GetStateHashCheckHeight(config.DefConfig.P2PNode.NetworkId)
+	if this.Blk.Header.Height >= stateHashHeight && this.MerkleRoot == common.UINT256_EMPTY {
+		return errors.New("received block msg with empty merkle root")
```
Honglei-Cong: It's not a deserialisation error.  It's better to move this check to block-msg-handler?

laizy: done

Honglei-Cong (approved):

PR #874, opened 2019-05-05
fix bug,when node slower than other node,save other node submit block…

xiemylogos:
… msg
@Honglei-Cong @laizy


consensus/vbft/service.go
```diff
@@ -1060,10 +1060,6 @@ func (self *Server) onConsensusMsg(peerIdx uint32, msg ConsensusMsg, msgHash com
 			return
 		}
 		msgBlkNum := pMsg.GetBlockNum()
-		if msgBlkNum > self.GetCurrentBlockNo() {
```
laizy, line 1063: no larger than 10 to avoid attack?

Honglei-Cong: before processing, consensus msg has been verified from valid consensus peer with signatures in msg.

Honglei-Cong:
From the log, suggest to change CAP_ACTION_CHANNEL from 8 to 64?

laizy (approved):

PR #916, opened 2019-05-23
add block height check when run transaction

carltraveler:


common/config/config.go
```diff
@@ -71,8 +71,9 @@ const (
 	DEFAULT_GAS_LIMIT                       = 20000
 	DEFAULT_GAS_PRICE                       = 500
 
-	DEFAULT_DATA_DIR      = "./Chain"
-	DEFAULT_RESERVED_FILE = "./peers.rsv"
+	DEFAULT_DATA_DIR                         = "./Chain"
+	DEFAULT_RESERVED_FILE                    = "./peers.rsv"
+	DEFAULT_OPCODE_HEIGH_UPDATE_FIRST uint32 = 4062597
```
laizy: need distinguish `testnet`, `mainnet` and `solo` mode, you can reference `STATE_HASH_HEIGHT_MAINNET`

Honglei-Cong:
It's better to set required-engine-version with smart-contract code.
required-engined-version is set when smart-contract is deployed.

When engine decoding smart-contract code, it will compare required-engine-version with op-code engine version.

common/constants/constants.go
```diff
@@ -82,3 +82,7 @@ const (
 // ledger state hash check height
 const STATE_HASH_HEIGHT_MAINNET = 3000000
 const STATE_HASH_HEIGHT_POLARIS = 850000
+
+// neovm opcode update check height
+const OPCODE_HEIGHT_UPDATE_FIRST_MAINNET = 4994530
```
laizy: how this value is calculated?

carltraveler: choose a future block about one month. any better idea?

laizy (dismissed):

laizy (approved):

PR #928, opened 2019-05-28
implement lock contract logic in sharding transaction

laizy:


Honglei-Cong:
need resolve conflicts.

PR #937, opened 2019-06-01
Removed unneeded spaces

WyattMufson:
Don't need the spaces for `[INFO]` and `[WARN]`

### DIFF
--- a/common/log/log.go
+++ b/common/log/log.go
@@ -59,8 +59,8 @@ const (
 var (
 	levels = map[int]string{
 		DebugLog: Color(Green, "[DEBUG]"),
-		InfoLog:  Color(Cyan, "[INFO ]"),
-		WarnLog:  Color(Yellow, "[WARN ]"),
+		InfoLog:  Color(Cyan, "[INFO]"),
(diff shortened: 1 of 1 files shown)


common/log/log.go
```diff
@@ -59,8 +59,8 @@ const (
 var (
 	levels = map[int]string{
 		DebugLog: Color(Green, "[DEBUG]"),
-		InfoLog:  Color(Cyan, "[INFO ]"),
-		WarnLog:  Color(Yellow, "[WARN ]"),
+		InfoLog:  Color(Cyan, "[INFO]"),
```
laizy, line 62: this is intended to make the ouput aligned，since `debug`,'error` 'fatal`,and `trace` are 5 letters.

WyattMufson: Is alignment required?

If so, would maybe adding a `.` or `!` be better than a space?

WyattMufson: It just looked a little off when I saw it

WyattMufson:
I think it is fine as is

PR #940, opened 2019-06-02
Fix logs

WyattMufson:
Fixed up spacing, capitalization and readability of some error logs

### DIFF
--- a/account/client.go
+++ b/account/client.go
@@ -120,7 +120,7 @@ func NewClientImpl(path string) (*ClientImpl, error) {
 func (this *ClientImpl) load() error {
 	err := this.walletData.Load(this.path)
 	if err != nil {
-		return fmt.Errorf("Error loading wallet %s: %s", this.path, err)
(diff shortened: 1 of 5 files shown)


account/client.go
```diff
@@ -318,7 +318,7 @@ func (this *ClientImpl) getAccount(accData *AccountData, passwd []byte) (*Accoun
 	addr := types.AddressFromPubKey(publicKey)
 	scheme, err := s.GetScheme(accData.SigSch)
 	if err != nil {
-		return nil, fmt.Errorf("signature scheme error:%s", err)
+		return nil, fmt.Errorf("Signature scheme error: %s", err)
```
laizy: we use lowercase to be consistent with golang std lib convention.

WyattMufson: Ah ok

WyattMufson: @laizy should be good now

laizy (approved):

PR #942, opened 2019-06-03
add wasm contract notify array support

JasonZhouPW:
add : when pass json format {"states":["aa","bb","cc"...]} to notify api in smart contract, node will send out the notify as same format as native/neovm contract.


smartcontract/service/wasmvm/runtime.go
```diff
@@ -156,7 +161,15 @@ func Notify(proc *exec.Process, ptr uint32, len uint32) {
...
 
-	notify := &event.NotifyEventInfo{self.Service.ContextRef.CurrentContext().ContractAddress, string(bs)}
+	var notify *event.NotifyEventInfo
+	nj := new(NotifyJson)
+	err = json.Unmarshal(bs, nj)
+	if err != nil {
+		notify = &event.NotifyEventInfo{self.Service.ContextRef.CurrentContext().ContractAddress, string(bs)}
+	} else {
+		notify = &event.NotifyEventInfo{self.Service.ContextRef.CurrentContext().ContractAddress, nj.States}
+	}
```
laizy, line 179: exact those code to a function like : 
```go
func ParseEventInfo(rawBytes []byte) *NotifyEventInfo;
```
and add unit test case for this function to document the codec convention.

JasonZhouPW:
deprecated

PR #954, opened 2019-06-18
consensus sharding sync and cross shard msg send

xiemylogos:
1、fix cross shard msg send
2、fix shard consensus cycle change
3、optimize shard call tx in p2p send
4、save some cross shard msg to db

### DIFF
--- a/consensus/vbft/node_utils.go
+++ b/consensus/vbft/node_utils.go
@@ -497,10 +497,10 @@ func (self *Server) SendCrossShardMsgToAll(height uint32) {
 		if err != nil {
 			if err != com.ErrNotFound {
 				log.Errorf("SendCrossShardMsgToAll getshardmsghash err:%s", err)
+				continue
 			}
-		} else {
-			crossShardMsg.CrossShardMsgInfo.PreCrossShardMsgHash = preMsgHash
 		}
+		crossShardMsg.CrossShardMsgInfo.PreCrossShardMsgHash = preMsgHash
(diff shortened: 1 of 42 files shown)


consensus/vbft/node_utils.go
```diff
@@ -497,9 +497,19 @@ func (self *Server) SendCrossShardMsgToAll(height uint32) {
 		if err != nil {
 			if err != com.ErrNotFound {
 				log.Errorf("SendCrossShardMsgToAll getshardmsghash err:%s", err)
+				continue
```
Honglei-Cong: continue? or return?

core/chainmgr/chainmgr.go
```diff
@@ -77,6 +76,7 @@ type ChainManager struct {
 	lock       sync.RWMutex
 	shards     map[common.ShardID]*ShardInfo
 	mainLedger *ledger.Ledger
+	ledger     *ledger.Ledger
```
Honglei-Cong: It's better to remove both mainLedger and ledger from chainmgr.
Getting ledger with ledger.GetShardLedger(...)?

consensus/vbft/node_utils.go
xiemylogos: maintain data consistency,use return

core/chainmgr/chainmgr.go
xiemylogos: ok

Honglei-Cong (approved):

PR #962, opened 2019-06-19
support shard block syncing

Honglei-Cong:


core/ledger/ledger.go
```diff
@@ -179,68 +178,81 @@ func (self *Ledger) AddHeaders(headers []*types.Header) error {
...
-			} else {
-				log.Errorf("Ledger ExecuteBlock GetBlock sharad height:%d,ParentHeight:%d error:%s", b.Header.Height, b.Header.ParentHeight, err)
+		currentParentHeight := self.ParentLedger.GetCurrentBlockHeight()
+		for ph := currentParentHeight + 1; ph <= b.Header.ParentHeight; ph++ {
+			parentBlock, merkleRoot, err := self.ParentBlockCache.GetBlock(ph)
+			if err != nil {
+				log.Errorf("Ledger %d ExecuteBlock GetBlock shard height:%d,ParentHeight:%d error:%s", self.ShardID, b.Header.Height, ph, err)
 				return store.ExecuteResult{}, err
 			}
+			result, err := self.ParentLedger.ldgStore.ExecuteBlock(parentBlock)
```
laizy: this will have problem, previous block result is not submitted, so will cause executing block at height `H+2` based on ledger state at `H`

Honglei-Cong: right.

PR #966, opened 2019-06-21
get shard config from parent ledger use memDB

xiemylogos:
1、shard chain  get shard config from parent ledger use memDB


laizy:
from the discussion， this will be simplified.

Honglei-Cong:
this has been fixed in another pr.

PR #1021, opened 2019-07-09
fix update shard config bug

qiluge:


core/store/ledgerstore/ledger_store.go
```diff
@@ -173,18 +173,20 @@ func (this *LedgerStoreImp) InitLedgerStoreWithGenesisBlock(genesisBlock *types.
...
-			return fmt.Errorf("HashBlockExist error %s", err)
-		}
-		if !exist {
-			return fmt.Errorf("GenesisBlock arenot init correctly, hash %s", genesisHash.ToHexString())
-		}
 		err = this.init()
 		if err != nil {
 			return fmt.Errorf("init error %s", err)
 		}
+		if this.currBlockHeight == 0 {
```
Honglei-Cong, line 180: remove this if?
the following checks are always needed.

core/chainmgr/chainmgr.go
```diff
@@ -196,7 +196,7 @@ func (self *ChainManager) initMainLedger(stateHashHeight uint32) error {
 }
 
 func (self *ChainManager) initShardLedger(shardInfo *ShardInfo) error {
-	if shardInfo != nil && ledger.GetShardLedger(shardInfo.ShardID) != nil {
+	if ledger.GetShardLedger(shardInfo.ShardID) != nil {
```
Honglei-Cong, line 199: necessary change here?

qiluge: If we don't remove it, we need to judge ```shardInfo == nil ``` in addition

core/store/ledgerstore/ledger_store.go
qiluge: This is a temporary change, which will be optimized later

PR #1046, opened 2019-08-05
fix wasm api memory attack

laizy:


smartcontract/service/wasmvm/storage.go
```diff
@@ -25,12 +25,23 @@ import (
...
+		return nil, errors.New("contract create len is greater than memory size")
+	}
+	keybytes := make([]byte, len)
+	_, err := proc.ReadAt(keybytes, int64(ptr))
+	if err != nil {
+		return nil, err
+	}
 
+	return keybytes, nil
+}
```
carltraveler: this function put it to runtime.go or wasm_service.go is better

laizy: extract out to utils.go

carltraveler (approved):

PR #1047, opened 2019-08-05
validate wasm before deploy

carltraveler:


laizy:
add rust validation code since https://github.com/ontio/wagon/pull/13 has merged.

laizy (approved):

PR #1060, opened 2019-08-26
add prefix ontio_ to runtime API name and add sha256 runtime API

carltraveler:


smartcontract/service/wasmvm/runtime.go
```diff
@@ -77,6 +78,22 @@ func SelfAddress(proc *exec.Process, dst uint32) {
 	}
 }
 
+func Sha256(proc *exec.Process, src uint32, len uint32, dst uint32) {
+	bs, err := ReadWasmMemory(proc, src, len)
```
laizy: missing check gas fee.

carltraveler: is that resolove later? like other api

smartcontract/service/wasmvm/runtime.go
```diff
@@ -513,125 +535,135 @@ func NewHostModule() *wasm.Module {
...
-			"current_txhash": {
-				FieldStr: "current_txhash",
+			"ontio_current_txhash": {
+				FieldStr: "ontio_current_txhash",
 				Kind:     wasm.ExternalFunction,
 				Index:    11,
 			},
-			"ret": {
-				FieldStr: "ret",
+			"ontio_ret": {
```
laizy: before we had to use `ret` instead `return` since `return` is rust/c++ keyword. but now we can rename to `ontio_return` to make it more clear.

laizy (approved):

PR #1062, opened 2019-08-28
add asset api support

tanZiWen:


smartcontract/service/neovm/config.go
```diff
@@ -104,6 +104,7 @@ var (
 	RUNTIME_ADDRESSTOBASE58_NAME     = "Ontology.Runtime.AddressToBase58"
 	RUNTIME_GETCURRENTBLOCKHASH_NAME = "Ontology.Runtime.GetCurrentBlockHash"
 	RUNTIME_VERIFYMUTISIG_NAME       = "Ontology.Runtime.VerifyMutiSig"
+	RUNTIME_ASSERT_NAME              = "Ontology.Runtime.Assert"
```
carltraveler: Assert api maybe confused with the python original keyword "assert".  suguest OntAssert

tanZiWen: ok

laizy:
breaking change need add effective height check for testnet and mainnet.

tanZiWen:
it's unnecessary to do that. it doesn't a function which has existed.

laizy:
I am afraid not, if compiler is ready, someone can deploy the code to testnet and mainnet at any time.

tanZiWen:
the compiler won't publish this function until this function has been launched on testnet and mainnet.

PR #1066, opened 2019-09-05
implement wasm test case framework

laizy:


core/store/ledgerstore/ledger_store.go
```diff
@@ -1028,10 +1028,15 @@ func (this *LedgerStoreImp) GetEventNotifyByBlock(height uint32) ([]*event.Execu
 //PreExecuteContract return the result of smart contract execution without commit to store
 func (this *LedgerStoreImp) PreExecuteContract(tx *types.Transaction) (*sstate.PreExecResult, error) {
 	height := this.GetCurrentBlockHeight()
+	// use previous block time to make it predictable for easy test
+	blockTime := uint32(time.Now().Unix())
+	if header, err := this.GetHeaderByHeight(height); err == nil {
+		blockTime = header.Timestamp + 1
+	}
```
carltraveler, line 1035: add err assert here?

wasmtest/run-wasm-tests.sh
```diff
@@ -0,0 +1,23 @@
...
+set -e
+set -x
+
+# install build tools
+if ! which rustup ; then
+	curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain nightly 
+	source $HOME/.cargo/env
+fi
+rustup target add wasm32-unknown-unknown
+which ontio-wasm-build || cargo install --git=https://github.com/ontio/ontio-wasm-build
```
carltraveler, line 11: this code setup move it to contracts-rust better?

laizy: `ontio-wasm-build` is used below, I prefer group the setup and usage   together.

core/store/ledgerstore/ledger_store.go
laizy: why need assert here? `err` could be not `nil`, in this case we fall back to use `time.Now`.

carltraveler (approved):

PR #1089, opened 2019-10-09
fix deploy code vm type compatibility

laizy:


core/payload/deploy_code.go
```diff
@@ -54,6 +65,30 @@ func (dc *DeployCode) Address() common.Address {
 	return dc.address
 }
 
+func (dc *DeployCode) SetVmType(ty VmType) {
+	dc.vmFlags = byte(ty)
+}
```
carltraveler, line 70: check ty here?

core/payload/deploy_code.go
```diff
@@ -252,10 +287,13 @@ func CreateDeployCode(code []byte,
 	author []byte,
 	email []byte,
 	desc []byte) (*DeployCode, error) {
+	if vmType > 255 {
+		return nil, fmt.Errorf("wrong vm flags: %d", vmType)
+	}
```
carltraveler, line 292: here is redundant. validateDeployCode already check

carltraveler (approved):

core/payload/deploy_code.go
laizy: this checks uint32 -> byte convension.

core/payload/deploy_code.go
laizy: ty is `VmType`, which is guranteed to be valid.

PR #1092, opened 2019-10-11
allow vm reader eof error to avoid hard-fork in testnet

laizy:


vm/neovm/utils/vm_reader.go
```diff
@@ -43,10 +44,16 @@ func (r *VmReader) ReadByte() (byte, error) {
...
+	amount := r.reader.Len()
+	if r.AllowEOF {
+		amount = 1 * 1024 * 1024
+	}
+	if amount < count {
 		return nil, io.EOF
 	}
 	b := make([]byte, count)
+	// when reader does not has enough bytes, it will return the actual length and error is nil
+	// this is a bug previously(it should use io.ReadFull), but we have to keep it to avoid hard fork
```
tanZiWen, line 56: why it will have a hard fork, please write down your reason instead of a title.

laizy: the reason is listed in line:55

carltraveler (approved):

PR #1106, opened 2019-10-17
Break whole loop in heartBeatService

rongyi:
The break statement's intention is to quit the for loop, but what it actually break is only the select block, i.e. we still in for loop, just because the main goroutine take down so this for loop goes down as well. Since there are no logic code after the for loop, a simple return will do.


laizy (approved):

laizy:
good catch

PR #1109, opened 2019-10-23
Change p2pserver connection cache from slice to set

rongyi:
1. the time complexity of searching in slice is O(n) so change from slice to map give a better performace.
2. ConnectingNodes field in NetServer need not to export, all access to this field is from the 4 API(AddOutConnectingList/RemoveFromConnectingList/GetOutConnectingListLen/ IsAddrFromConnecting) so lowercase this field.
3. the other two cache(inConnRecord/outConnRecord) are similar with ConnectingNodes
4. Add testcase to cover those change.

### DIFF
--- a/p2pserver/common/set/string_set.go
+++ b/p2pserver/common/set/string_set.go
@@ -1,3 +1,21 @@
+/*
+ * Copyright (C) 2018 The ontology Authors
+ * This file is part of The ontology library.
+ *
+ * The ontology is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * The ontology is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with The ontology.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
 package set
 
 import (
--- a/p2pserver/common/set/string_set.go
+++ b/p2pserver/common/set/string_set.go
@@ -1,0 +1,189 @@
+package set
+
+import (
+	"reflect"
+	"sort"
+)
+
+type Empty struct{}
+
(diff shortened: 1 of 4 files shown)


p2pserver/common/set/string_set.go
```diff
@@ -0,0 +1,189 @@
+package set
+
+import (
+	"reflect"
+	"sort"
+)
+
+type Empty struct{}
```
laizy: internal type should not be public.

p2pserver/common/set/string_set.go
```diff
@@ -0,0 +1,189 @@
...
+
+// sets.String is a set of strings, implemented via map[string]struct{} for minimal memory consumption.
+type String map[string]Empty
```
laizy: It's an uncomfortable name. I prefer `Strings` or `StringSet`

p2pserver/common/set/string_set.go
```diff
@@ -0,0 +1,189 @@
...
+// NewString creates a String from a list of values.
+func NewString(items ...string) String {
+	ss := String{}
+	ss.Insert(items...)
+	return ss
+}
+
+// StringKeySet creates a String from a keys of a map[string](? extends interface{}).
+// If the value passed in is not actually a map, this will panic.
+func StringKeySet(theMap interface{}) String {
```
laizy: remove all unused functions

p2pserver/common/set/string_set.go
```diff
@@ -0,0 +1,189 @@
...
+	return false
+}
+
+// Difference returns a set of objects that are not in s2
+// For example:
+// s1 = {a1, a2, a3}
+// s2 = {a1, a2, a4, a5}
+// s1.Difference(s2) = {a3}
+// s2.Difference(s1) = {a4, a5}
+func (s String) Difference(s2 String) String {
```
laizy: I notice slightly inconsistent semantics: some function like `Delete` and `Insert` modify the original set and return it. but this function and the following one keep the original set unchanged and return a new one. I suggest removing the return value from `Delete` and `Insert` to avoid misuse. 
add unit test or remove this function since it is not used.

p2pserver/common/set/string_set.go
rongyi: Thanks for pointing out

p2pserver/common/set/string_set.go
rongyi: Done

p2pserver/common/set/string_set.go
rongyi: Remove unused API

laizy (approved):

PR #1121, opened 2019-11-04
add Proposal as Endorse-sig

Honglei-Cong:
fix f faulty-node issue.

1. add proposer's signature to endorse-signature-collection
2. continue processing proposal/endorsement-msg even after node is in phase COMMIT

### DIFF
--- a/consensus/vbft/block_pool.go
+++ b/consensus/vbft/block_pool.go
@@ -147,6 +147,15 @@ func (pool *BlockPool) newBlockProposal(msg *blockProposalMsg) error {
 
 	// add msg to proposals
 	candidate.Proposals = append(candidate.Proposals, msg)
+
+	// add endorse-sig
+	proposer := msg.Block.getProposer()
+	eSig := &CandidateEndorseSigInfo{
+		EndorsedProposer: proposer,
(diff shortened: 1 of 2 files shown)


consensus/vbft/block_pool.go
```diff
@@ -147,6 +147,15 @@ func (pool *BlockPool) newBlockProposal(msg *blockProposalMsg) error {
 
 	// add msg to proposals
 	candidate.Proposals = append(candidate.Proposals, msg)
+
+	// add endorse-sig
+	proposer := msg.Block.getProposer()
+	eSig := &CandidateEndorseSigInfo{
+		EndorsedProposer: proposer,
+		Signature:        msg.Block.Block.Header.SigData[0],
```
laizy, line 155: where is the code that guarantee the proposer's signature is the first one  when msg comes from faulty node?

Honglei-Cong: 2f+1 voting, should include proposer.
proposal msg dont need to be the first one.

Honglei-Cong: https://github.com/ontio/ontology/blob/master/consensus/vbft/msg_types.go#L72

laizy (approved):

PR #1125, opened 2019-11-11
Add some small improvement

rongyi:
1. in main.go:initRpc()/initLocalRpc() flag is useless

```
	flag := false
	select {
	case <-exitCh:
		if !flag {
			return err
		}
	case <-time.After(time.Millisecond * 5):
		flag = true
	}
```

  select is a one time statement. if select
  + goes to the first case, flag is definitely false, so the check is redundant. 
  + goes to the second case, flag is not used after the setting. 
so there's no need to add this flag variable.
2. some inner function should not be exported: NbrPeers::NodeExisted
3. ` if p == nil || (p != nil && p.GetState() != common.ESTABLISH)`  when the or statement goes to the second part it means the first part is already evaluated to false, in this case it means p == nil evaluated to false so p is not nil. so the second part p != nil is included implicitly. Drop this check in second part.
4. change some benchmark test to golang Benchmark testing
5. imrove nbrpeers test case.
6. typo in test
7. unnecessary guard around call to delete, Deleting a non-existent entry in map is a no-op.

### DIFF
--- a/main.go
+++ b/main.go
@@ -344,23 +344,23 @@ func initRpc(ctx *cli.Context) error {
 		return nil
 	}
 	var err error
-	exitCh := make(chan interface{}, 0)
+	exitCh := make(chan struct{}, 0)
 	go func() {
 		err = jsonrpc.StartRPCServer()
 		close(exitCh)
 	}()
 
-	flag := false
+	// StartRPCServer will call http.ListenAndServe
+	// so when return immediately means it does not start OK
+	// regist 30 handler and call http.ListenAndServe consume
+	// 150us approximately, 5ms here is enough
 	select {
 	case <-exitCh:
-		if !flag {
-			return err
-		}
+		return err
 	case <-time.After(time.Millisecond * 5):
-		flag = true
+		log.Infof("Rpc init success")
+		return nil
 	}
-	log.Infof("Rpc init success")
-	return nil
 }
 
 func initLocalRpc(ctx *cli.Context) error {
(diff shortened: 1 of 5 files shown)


qiluge (approved):

main.go
```diff
@@ -344,48 +344,47 @@ func initRpc(ctx *cli.Context) error {
 		return nil
 	}
 	var err error
-	exitCh := make(chan interface{}, 0)
+	exitCh := make(chan struct{}, 0)
```
Honglei-Cong, line 347: why not change exitCh to errCh, and send err back in channel

p2pserver/message/types/message_test.go
```diff
@@ -61,27 +60,26 @@ func TestMsgHdr2(t *testing.T) {
 	if hdr1.Length != hdr.Length ||
 		hdr1.Magic != hdr.Magic ||
 		hdr1.CMD != hdr.CMD ||
-		hdr1.Checksum != hdr1.Checksum {
```
Honglei-Cong, line 64: o(╥﹏╥)o

main.go
```diff
@@ -344,48 +344,47 @@ func initRpc(ctx *cli.Context) error {
...
-	return nil
 }
 
 func initLocalRpc(ctx *cli.Context) error {
 	if !ctx.GlobalBool(utils.GetFlagName(utils.RPCLocalEnableFlag)) {
 		return nil
 	}
 	var err error
-	exitCh := make(chan interface{}, 0)
+	exitCh := make(chan struct{}, 0)
```
Honglei-Cong, line 371: ditto
change to `chan error`?

PR #1152, opened 2020-02-17
fix ontid group toJson bug

lucas7788:


Honglei-Cong:
In `regIdWithController`, serialized `group` is provided in args.  This means SDK support `group` serialization/deserialization.
IMHO, `getDDO` can just return serialized `group`.

What you think? @AlverLyu

AlverLyu:
> In `regIdWithController`, serialized `group` is provided in args. This means SDK support `group` serialization/deserialization.
> IMHO, `getDDO` can just return serialized `group`.
> 
> What you think? @AlverLyu

This is not a big problem.

IMO, it is better to use json in all interfaces, since it is more friendly. The main reason for using the ontology serialization in the 'set' interfaces is to reduce the data size, while this is not a critical problem in 'get' interfaces.

Anyway, either is OK.

siovanus:
This issue is done by https://github.com/ontio/ontology/pull/1181

PR #1155, opened 2020-02-24
drop block if catch internal err

carltraveler:


smartcontract/smart_contract.go
```diff
@@ -51,6 +51,7 @@ type SmartContract struct {
 	WasmExecStep  uint64
 	JitMode       bool
 	PreExec       bool
+	InternelErr   bool
```
laizy: why public here since we have set and get method?

laizy (approved):

PR #1156, opened 2020-02-25
fix cmd pre bug

lucas7788:


laizy:
the change is not obvious enough, need add commit log

laizy (approved):

PR #1173, opened 2020-03-31
fix consensus cross chain msg compatible old version

tanZiWen:


consensus/vbft/types.go
```diff
@@ -135,8 +135,8 @@ func (blk *Block) Deserialize(data []byte) error {
 	}
 
 	var crossChainMsg *types.CrossChainMsg
-	hasEmptyCCM, irr, eof := source.NextBool()
-	if irr || eof {
+	hasEmptyCCM, irr, _ := source.NextBool()
```
laizy, line 139: document why not handle eof here

tanZiWen: added

consensus/vbft/block_pool.go
```diff
@@ -477,6 +477,7 @@ func (pool *BlockPool) newBlockCommitment(msg *blockCommitMsg) error {
 		if crossChainMsgSig, present := msg.CrossChainMsgEndorserSig[endorser]; present {
```
Honglei-Cong: check if CrossChainMsgEndorserSig is nil?

tanZiWen: msg_builder // 383 had made CrossChainMsgEndorserSig

Honglei-Cong (approved):

PR #1180, opened 2020-04-28
support async send with buffer size limit

laizy:


p2pserver/link/link.go
```diff
@@ -190,20 +258,66 @@ func (this *Link) needSendMsg(msg types.Message) bool {
...
+					if err != nil {
+						return
+					}
+					buffers = buffers[:0]
+					results = results[:0]
+				}
+			}
+		} else {
+			// no buffer has been taken, yield this goroutine to avoid busy loop
+			runtime.Gosched()
```
Honglei-Cong, line 312: can this be implemented with channel?

laizy: It will cause sending a notification for every msg to wake the writer, which will hurt performance in high contention scenario.

PR #1200, opened 2020-05-14
fix testmode start bug

lucas7788:


laizy (approved):

laizy:
use more detailed discription next time, instead of `fix bug`.

PR #1201, opened 2020-05-14
check peer existence before reconnecting.

laizy:
when two nodes established 2 connection, the first one will be closed.
but the address will be added in reconnect list, and the reconnect rely
on NodeConnect system message to remove the node. so it will never be removed,
since the NodeConnect message of the second connection has already been dispatched.

### DIFF
--- a/p2pserver/protocols/reconnect/reconnect.go
+++ b/p2pserver/protocols/reconnect/reconnect.go
@@ -30,8 +30,8 @@ import (
 	"github.com/ontio/ontology/p2pserver/peer"
 )
 
-type Info struct {
-	count int
+type ReconnectPeerInfo struct {
+	count int // current retry count
 	id    common.PeerId
 }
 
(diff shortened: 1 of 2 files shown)


p2pserver/protocols/reconnect/reconnect.go
```diff
@@ -32,11 +30,16 @@ import (
...
+	count int
+	id    common.PeerId
+}
+
 //ReconnectService contain addr need to reconnect
 type ReconnectService struct {
 	sync.RWMutex
 	MaxRetryCount uint
-	RetryAddrs    map[string]int
+	RetryAddrs    map[string]Info
```
Honglei-Cong:
```suggestion
	RetryAddrs    map[string]*Info
```

p2pserver/protocols/reconnect/reconnect.go
```diff
@@ -32,11 +30,16 @@ import (
 	"github.com/ontio/ontology/p2pserver/peer"
 )
 
+type Info struct {
```
Honglei-Cong:
```suggestion
type ReconnectPeerInfo struct {
```

Honglei-Cong (approved):

PR #1226, opened 2020-06-11
add query api for newo vm

siovanus:
add query api for newo vm


qiluge:
maybe the title `newo vm` not correct.

Honglei-Cong (approved):

PR #1249, opened 2020-07-05
support v2 version in go mod

xiemylogos:
ontology release v2.0.0 version ,then go mod support v2 and beyond.
reference:https://blog.golang.org/v2-go-modules,https://github.com/golang/go/issues/35732


Honglei-Cong:
Why not create a new repo?

laizy:
have created v1.11.0 tag for lib dependence, and take v2.0.0 as executable version.

PR #1259, opened 2020-07-10
Fix data race

Honglei-Cong:


consensus/vbft/block_pool.go
```diff
@@ -631,15 +631,22 @@ func (pool *BlockPool) addSignaturesToBlockLocked(block *Block, forEmpty bool) e
 			}
 		}
 	}
+
+	// clone a new block
+	newBlk := &Block{}
+	if err := newBlk.Deserialize(block.Serialize()); err != nil {
```
laizy: this looks too hacky, adding a new proposalSig field in blockProposalMsg to avoid data race.

Honglei-Cong: Right.  But block signatures needs to be updated, and there'll be data race anyway.
You'll have to add lock to block header, or copy a new one.

laizy: this update function is called when the blockPool's lock holds, I do not expect there is any data race left.

Honglei-Cong: No lock when reading signatures in block header, but signature array in block header must be updated.
Either adding locks for all read operations, or just clone a new one when updating then there'll no write ops to block header?

Honglei-Cong: blockpool's lock can not protect block headers in proposal message?  

ヽ༼ຈʖ̯ຈ༽ﾉ
--

Honglei-Cong: updated

laizy (approved):

PR #1263, opened 2020-07-21
fix log data race

laizy:


common/log/log.go
```diff
@@ -85,7 +87,19 @@ func GetGID() uint64 {
 	return n
 }
 
-var Log *Logger
+var globalLogger unsafe.Pointer
```
Honglei-Cong, line 90: validate pointer must not be nil after init?

common/log/log.go
```diff
@@ -368,11 +369,20 @@ func InitLog(logLevel int, a ...interface{}) {
...
+
+	logger := New(fileAndStdoutWrite, "", log.LUTC|log.Ldate|log.Lmicroseconds, logLevel, logFile)
+
+	return logger
+}
+
+func InitLog(logLevel int, a ...interface{}) {
+	logger := createLog(logLevel, a...)
+
+	swapGlobalLogger(logger)
```
Honglei-Cong: It's public.
check if old-logger is nil, close it if not nil?

common/log/log.go
```diff
@@ -85,7 +87,19 @@ func GetGID() uint64 {
 	return n
 }
 
-var Log *Logger
+var globalLogger unsafe.Pointer
+
+func Log() *Logger {
```
Honglei-Cong, line 92: just a comment.
this function seems not necessary to be public.

Honglei-Cong:
looks good to me.

common/log/log.go
laizy: used in http rpc to set the log level.

common/log/log.go
laizy: done

common/log/log.go
laizy: there is a init function to call InitLog, so can gurantee globalLogger will never be nil.

Honglei-Cong (approved):

PR #1284, opened 2020-10-20
fix for consensus

zhiqiangxu:


consensus/vbft/state_mgmt.go
```diff
@@ -385,41 +385,48 @@ func (self *StateMgr) checkStartSyncing(startBlkNum uint32, forceSync bool) {
...
 	}
 
+	var maxCommittedPeers []uint32
+	if maxCommitted > 0 {
+		for _, k := range self.peers {
+			if k.committedBlockNum >= maxCommitted {
+				maxCommittedPeers = append(maxCommittedPeers, k.peerIdx)
+			}
+		}
+	}
```
laizy, line 412: move into the following if  block

laizy, line 412: move into the following if block?

xiemylogos (approved):

PR #1298, opened 2021-01-21
optimize sending large packet

zhiqiangxu:


p2pserver/link/link.go
```diff
@@ -152,21 +152,32 @@ func (this *Link) SendRaw(rawPacket []byte) error {
...
-		return err
-	}
 
-	return nil
+	start := time.Now()
+	deadline := start.Add(time.Duration(nCount*common.WRITE_DEADLINE) * time.Second)
+	wrote := 0
+	for {
+		conn.SetWriteDeadline(deadline)
+		n, err := conn.Write(rawPacket[wrote:])
```
laizy, line 171: writing partial packet will cause data corruption, because of data race.

zhiqiangxu: added lock protection.

PR #1308, opened 2021-05-19
Fix decoding of event state without loss of numeric precision

tav:
When decoding events such as ONG transfers, numeric values are decoded to `float64` by the default `encoding/json` configuration. This results in the loss of precision for large integer values which cannot be accurately represented by a `float64`.

This PR enables `UseNumber` on the JSON decoder so that consumers can convert the resulting `json.Number` appropriately, without having to lose any precision.


laizy (approved):
thanks.

PR #1321, opened 2021-06-03
clean validator actor

laizy:


txnpool/proc/txnpool_worker.go
```diff
@@ -62,6 +85,8 @@ func NewTxPoolWoker(s *TXPoolServer) *txPoolWorker {
 	worker.rspCh = make(chan *types.CheckResponse, tc.MAX_PENDING_TXN)
 	worker.stopCh = make(chan bool)
 	worker.server = s
+	worker.stateless = stateless.NewValidatorPool(2)
```
rongyi, line 88: May be worker can be more than 2? workerpool will shrink when idle in a specified period.

txnpool/proc/txnpool_actor.go
```diff
@@ -119,71 +114,57 @@ func (ta *TxActor) handleTransaction(sender tc.SenderType, self *actor.PID,
...
+		if ok, desc := preExecCheck(txn); !ok {
+			log.Debugf("handleTransaction: preExecCheck tx %x failed", txn.Hash())
+			replyTxResult(sender, txResultCh, txn.Hash(), errors.ErrUnknown, desc)
+			return
 		}
-		<-ta.server.slots
-		ta.server.assignTxToWorker(txn, sender, txResultCh)
+		log.Debugf("handleTransaction: preExecCheck tx %x passed", txn.Hash())
 	}
+	<-ta.server.slots
```
rongyi, line 163: What if we drop the slots? let the worker handle what it can handle at most.

txnpool/proc/txnpool_server.go
```diff
@@ -307,17 +285,9 @@ func (s *TXPoolServer) setPendingTx(tx *tx.Transaction,
...
 
 // assignTxToWorker assigns a new transaction to a worker by LB
-func (s *TXPoolServer) assignTxToWorker(tx *tx.Transaction,
-	sender tc.SenderType, txResultCh chan *tc.TxResult) bool {
-
-	if tx == nil {
-		return false
-	}
-
+func (s *TXPoolServer) assignTxToWorker(tx *tx.Transaction, sender tc.SenderType, txResultCh chan *tc.TxResult) bool {
```
rongyi, line 283: return bool value is neve used

txnpool/proc/txnpool_worker.go
```diff
@@ -176,23 +164,15 @@ func (worker *txPoolWorker) verifyTx(tx *tx.Transaction) {
...
+		log.Debugf("verifyTx: transaction %x already in the verifying process", tx.Hash())
 		return
 	}
-	// Construct the request and send it to each validator server to verify
-	req := &types.CheckTx{
-		Tx: tx,
-	}
 
-	worker.sendReq2Validator(req)
+	worker.startFullVerify(tx)
```
rongyi: I think safe order is put to map first and then start verify, in case when verify process is so fast, the comming back response will not found in pendingTxList

PR #1323, opened 2021-06-08
simplify block verification

laizy:


txnpool/proc/txnpool_server.go
```diff
@@ -483,63 +397,90 @@ func (s *TXPoolServer) verifyBlock(req *tc.VerifyBlockReq, sender *actor.PID) {
...
+		for currHeight < req.Height {
+			// wait ledger sync up
+			log.Warnf("ledger need sync up for tx verification, curr height: %d, expected:%d", currHeight, req.Height)
+			time.Sleep(time.Second)
+			currHeight = ledger.DefLedger.GetCurrentBlockHeight()
+		}
 
-	for _, t := range checkBlkResult.VerifiedTxs {
-		s.pendingBlock.processedTxs[t.Tx.Hash()] = t
+		ch := make(chan *types.CheckResponse, len(checkBlkResult.UnverifiedTxs))
```
rongyi: channel size should be lenStateFul?

rongyi:
we return VerifyBlockRsp asap when in error, so in response process we can iterate the list in reverse order to get error asap(or just check the last one) Other than that LGTM

PR #1324, opened 2021-06-09
clean tx pool worker

laizy:


rongyi:
LGTM
other than some variable name,  txnpool now looks so clean!

PR #1338, opened 2021-06-17
add expiration check of eip155 tx in tx pool

JasonZhouPW:
1. check will triggered every 30 sec by default
2. if pending eiptx count exceed 500
3. remove the submitted time passed 10 minutes eip155 pending tx in tx pool

### DIFF
--- a/common/config/config.go
+++ b/common/config/config.go
@@ -25,6 +25,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"time"
 
 	"github.com/ontio/ontology-crypto/keypair"
 	"github.com/ontio/ontology/common"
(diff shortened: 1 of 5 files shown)


common/config/config.go
```diff
@@ -84,6 +85,12 @@ const (
 
 	//DEFAULT_ETH_BLOCK_GAS_LIMIT = 800000000
 	DEFAULT_ETH_TX_MAX_GAS_LIMIT = 6000000
+
+	//ticker interval to clear check eip tx in seconds
+	CLEAR_EIPTX_INTERVAL = 30 * time.Second
+	//eip tx expiration time in seconds
+	EIPTX_EXPIRATION_TIME = 10 * 60
+	EIPTX_POOL_MAX_COUNT  = 500
```
laizy, line 93: 这种纯实现细节，其他模块压根不会用到的参数挪到对应的模块里去。

PR #1340, opened 2021-06-17
add pending eiptx expiration check

JasonZhouPW:


txnpool/common/transaction_pool.go
```diff
@@ -155,14 +179,17 @@ func (tp *TXPool) AddTxList(txEntry *VerifiedTx) errors.ErrCode {
...
 	var cleaned []*types.Transaction
 	for _, tx := range txs {
 		if tx.IsEipTx() {
 			if _, ok := s.eipTxPool[tx.Payer]; ok {
 				cleaned = append(cleaned, s.eipTxPool[tx.Payer].Forward(uint64(tx.Nonce+1))...)
 				if s.eipTxPool[tx.Payer].Len() == 0 {
 					delete(s.eipTxPool, tx.Payer)
+					delete(s.userLatestEiptxHeight, tx.Payer)
+				} else {
+					s.userLatestEiptxHeight[tx.Payer] = height
```
rongyi, line 192: if eipTxPool has a malicious tx(say valid nonce is 1, and the tx nonce is 1000), then we update the height and we lost the track for this malicious tx. right?

PR #1373, opened 2021-09-18
Upgrade crypto version

rongyi:


angelo1313aguilar:
We can't do any transfers on this date, j was trying to either transfer deposit or withdraw but o could NOT.

laizy:
@angelo1313aguilar Hi, can you describe the issue more clearly? Does the issue have any relationship with this PR?

PR #1375, opened 2021-10-09
fix EVM memory correction bug

laizy:
https://github.com/ethereum/go-ethereum/security/advisories/GHSA-9856-9gg9-qcmq


diucicd (commented):
LGTM

PR #1391, opened 2022-02-24
allow preexecute to call v2 read only method of ont/ong

laizy:
v2 balance of function is called by restful api internally


Karldodz29 (approved):

blue-valley:
sigtaransfertxv2

PR #1429, opened 2023-03-31
pin rust nightly version

laizy:


laizy:
new nightly version will generate wasm binary with post-MVP version instructions.

